PR #4, opened 2020-02-18
oc login before each run

Karm:
Enables the job parameter to choose the cluster.


Ladicek:
I don't know why, but indentation is wrong, otherwise -- yea, we should do this.

Karm:
@Ladicek Fixed.

PR #7, opened 2020-02-19
ephemeral namespaces & retaining resources after failure

### LINKED ISSUE #5: If oc new-project was not run, TS fails ConfigMap on HTTP 503
It is apparently required to have a project created prior to running the TS, i.e. ```oc login``` is not enough on its own, one has to do ```oc new-project ``` too.

I suggest we parametrize the project name and let the TS create and delete the project.

Ladicek:
Resolves #5.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,12 +9,12 @@ pipeline {
     stages {
         stage('Login') {
             steps {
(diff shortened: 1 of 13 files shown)


Ladicek:
I think you can just read the `README.md` change, to understand how this is supposed to be used. But if you review the code as well, I won't object :-)

PR #8, opened 2020-02-21
GitHub Actions to run surefire tests on merge / PR

rsvoboda:
GitHub Actions to run surefire tests on merge / PR

failsafe (OpenShift cluster) is out of scope of this PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,32 @@
+name: "Pull Request / Push Build"
+on:
+  - push
+  - pull_request
+jobs:
+  build:
+    name: JVM build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 11 ]
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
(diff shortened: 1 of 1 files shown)


rsvoboda:
@Ladicek can you review pls ?

Ladicek:
Ah, one idea: I don't know if it's possible in GitHub Actions, but I've seen others running a temporary OpenShift cluster in CI via `oc cluster up` (which is of course only available for OpenShift 3). On the other hand, the `*OpenShiftIT` tests here are meant to be executed against a real-world-ish cluster, and while the existing ones are small, that doesn't have to be the case all the time. So perhaps we should stick with Surefire-only anyway.

rsvoboda:
crap, I see [ERROR] Non-resolvable import POM: Could not find artifact io.quarkus:quarkus-bom:pom:999-SNAPSHOT @ line 40, column 25

https://github.com/rsvoboda/quarkus-openshift-test-suite/commit/f8ceebd66321339e0ed9b964c6ee9861f23be372/checks?check_suite_id=473332581

Ladicek:
Just wanted to comment about this exact thing. Are we ready for stabilizing on 1.2.1.Final? I'm afraid it doesn't have the fixes for the Kubernetes extension.

rsvoboda:
I will update GH action to build quarkus master

PR #20, opened 2020-04-24
Info about Quarkus 1.3 branch fail for MP module with @Asynchronous

rsvoboda:
Info about Quarkus 1.3 branch fail for MP module with @Asynchronous


Ladicek:
I think we should rather create a `1.3.x` branch, which wouldn't include the MP module at all.

Ladicek:
https://github.com/quarkus-qe/quarkus-openshift-test-suite/commit/423481994c5c4fa636d4df331dc7056f3a8cf3f3

PR #24, opened 2020-04-28
Sort and remove unused imports during the build

rsvoboda:
Sort and remove unused imports during the build

Look at pom.xml changes, the rest was produced by `mvn clean install -DskipTests -DskipITs`


Ladicek:
This is good, but the style totally has to match what IntelliJ does :-) Unfortunately, it isn't 100% possible, but it's easy to get very close. I'll push to this PR and merge.

rsvoboda:
Quarkus is using "wrong" setup too ?

https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L379-L389

Ladicek:
Yes.

Ladicek:
Thanks! :-)

rsvoboda:
Welcome ;)

PR #27, opened 2020-05-04
add messaging test with ActiveMQ Artemis and the Artemis JMS Quarkus extension

Ladicek:


messaging/artemis/src/test/resources/amq.yaml
```diff
@@ -0,0 +1,171 @@
...
+  metadata:
+    name: amq-broker
+  spec:
+    lookupPolicy:
+      local: false
+    tags:
+    - name: latest
+      from:
+        kind: DockerImage
+        name: registry.access.redhat.com/amq-broker-7/amq-broker-72-openshift
```
rsvoboda, line 25: What if I want AMQ 7.6? I can just rewrite it here I guess.

Ladicek: 7.6 is behind auth, so I went with 7.2, the latest that's accessible without auth. Changing this single line should be enough, yea.

rsvoboda: Thanks for confirmation.

PR #31, opened 2020-05-12
 Making sure that the test suite's OpenShift client trusts certificates provided by API server explicitely.

mjurc:


common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftClientResource.java
```diff
@@ -2,13 +2,18 @@
...
+import io.fabric8.openshift.client.OpenShiftConfigBuilder;
 import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
 
 final class OpenShiftClientResource implements CloseableResource {
     final OpenShiftClient client;
 
     static OpenShiftClientResource createDefault() {
-        return new OpenShiftClientResource(new DefaultOpenShiftClient());
+        OpenShiftConfig config = new OpenShiftConfigBuilder()
+                .withTrustCerts(true)
```
rsvoboda, line 14: So this is equivalent to `quarkus.kubernetes-client.trust-certs=true`, right?

Ladicek: Yes.

mjurc: Indeed:

https://github.com/quarkusio/quarkus/blob/53d41bd92b127d3ecc76f9678414343696ded7c8/extensions/kubernetes-client/runtime/src/main/java/io/quarkus/kubernetes/client/runtime/KubernetesClientUtils.java#L53

Ladicek: I mean, the commit message is misleading :-)

rsvoboda: Should @mjurc amend it?

mjurc: I've updated it :)

PR #32, opened 2020-05-14
Adds AMQP messaging smoke test

Karm:
AMQP smoke test with a tiny speck of reactive messaging...

### DIFF
--- a/messaging/amqp-reactive/.dockerignore
+++ b/messaging/amqp-reactive/.dockerignore
@@ -1,0 +1,4 @@
+*
+!target/*-runner
+!target/*-runner.jar
+!target/lib/*
--- a/messaging/amqp-reactive/pom.xml
+++ b/messaging/amqp-reactive/pom.xml
@@ -1,0 +1,89 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.quarkus.ts.openshift</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+
+    <artifactId>messaging-amqp</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Quarkus OpenShift TS: Messaging: AMQP</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-reactive-messaging-amqp</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-openshift</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus.ts.openshift</groupId>
+            <artifactId>app-metadata</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus.ts.openshift</groupId>
+            <artifactId>common</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-amqp-protocol</artifactId>
+            <version>${version.artemis.server}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-server</artifactId>
+            <version>${version.artemis.server}</version>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jboss.logmanager</groupId>
+                    <artifactId>jboss-logmanager</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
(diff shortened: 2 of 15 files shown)


messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceProducer.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private final Integer[] ints = new Integer[]{10, 20, 30, 40, 50, 60, 70, 80, 90, 100};
+    private final AtomicInteger p = new AtomicInteger(0);
+
+    private static final Logger LOG = Logger.getLogger(PriceProducer.class.getName());
+
+    @Outgoing("generated-price")
+    public Flowable<Integer> generate() {
+        LOG.info("generate fired...");
+        return Flowable.interval(1, TimeUnit.SECONDS)
```
Ladicek, line 23: RxJava vs. Mutiny?

Ladicek:
Do I understand correctly that the decision was to use Reactive Messaging instead of the QPid JMS client?

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceConsumer.java
```diff
@@ -0,0 +1,28 @@
...
+    private final List<Integer> prices = Collections.synchronizedList(new ArrayList<>());
+
+    @Incoming("prices")
+    public void process(Integer price) {
+        LOG.info("process fired: " + price);
+        this.prices.add(price);
+    }
+
+    public List<Integer> getPrices() {
+        return prices;
```
Ladicek, line 26: Do you know what bug happens here?

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceProducer.java
```diff
@@ -0,0 +1,33 @@
...
+                .map(tick -> next());
```
Ladicek, line 24: Would it be easiest to do `map(tick -> ((tick * 10) % 100) + 10)` instead of the whole dance with `AtomicInteger`?

Ladicek: Or `intervalRange` if we don't want to send messages forever.

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceProducer.java
Karm: @Ladicek  I don't know.  It seems Vert.x examples vs. Quarkus examples.    I suppose Mutiny then should be it?

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceConsumer.java
Karm: @Ladicek Hmm...it probably should give away the reference to the original list and it should create a new one to return? Not sure.

Karm:
> Do I understand correctly that the decision was to use Reactive Messaging instead of the QPid JMS client?

I don't know.  @rsvoboda asked me to look at https://github.com/amqphub/quarkus-qpid-jms-quickstart, I naturally also looked here https://github.com/quarkusio/quarkus-quickstarts/tree/master/amqp-quickstart  

This is what came out of it. I haven't worked with any of these APIs before, so if you have any pointers as to what actually makes sense to have in OpenShift TS, it is perhaps the time to summarize them here.

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceProducer.java
Karm: @Ladicek Addressed in the latest commit.

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceConsumer.java
Ladicek: `prices` is `Collections.synchronizedList`, so iteration over it must also be synchronized. You can either return a defensive copy here, or get rid of `synchronizedList` and use something like `ConcurrentLinkedQueue` or `CopyOnWriteArrayList`.

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceProducer.java
Ladicek: Probably.

Ladicek:
I think having a test for Reactive Messaging is also valuable, so I wouldn't delete it. But we should probably also have a test for QPid JMS.

messaging/amqp/src/main/java/io/quarkus/ts/openshift/messaging/amqp/PriceConsumer.java
Karm: @Ladicek Thanks. The latest commit addresses that.

rsvoboda:
@Karm , Trello card says `AMQP 1.0 JMS client - Apache Qpid JMS` and Qpid, the intention was to get coverage for `org.amqphub.quarkus:quarkus-qpid-jms`.

Let's keep this Reactive Messaging one, just rename messaging/amqp =>  messaging/amqp-reactive and this PR can be merged.

Can you introduce new `messaging/amqp-qpid-jms` PR for `org.amqphub.quarkus:quarkus-qpid-jms` during tomorrow? https://quarkus.io/guides/jms#qpid-jms-amqp uses similar PriceResource as you have in this PR, so the switch could be relatively easy.

Karm:
@rsvoboda  Renamed and squashed. Tests passed again.  Can be merged now.

Karm:
@rsvoboda @Ladicek  O.K., as soon as you merge this #32 , there is another one adding QPID https://github.com/Karm/quarkus-openshift-test-suite/tree/qpid  

It is actually a copy-paste of messaging/artemis Java code and messaging/amqp-reactive broker configuration.

Someone might come one day and unify these test apps, if needed.  

As I gradually learned about QPID's/AMQP existence, it becomes apparent that the whole thing might have been structured in a much simpler way.

rsvoboda:
> Someone might come one day and unify these test apps, if needed.

yes, I'm aware of this. We are also duplicating yaml files, we will need to improve this

Ladicek:
For the YAML files, it should fortunately be pretty easy -- they are loaded from classpath (`ClassLoader.getResource()` or something like that, writing from memory), so they can be easily moved out to a single common artifact that would be added as a `<scope>test</scope>` dependency and nothing else would change.

Ladicek:
When it comes to unifying the test apps, there's a precedent in `sql-db`.

Ladicek:
Also the supporting classes (e.g. the Artemis server) could be moved to a single artifact that would be added as a dependency.

PR #38, opened 2020-05-17
Adding basic scaling tests.

mjurc:


rsvoboda:
```
 [ERROR] /home/runner/work/quarkus-openshift-test-suite/quarkus-openshift-test-suite/scaling/src/main/java/io/quarkus/ts/openshift/scaling/ScalingResource.java:[4,19] package javax.ws.rs does not exist
```

rsvoboda:
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 104.179 s <<< FAILURE! - in io.quarkus.ts.openshift.scaling.ScalingOpenShiftIT
[ERROR] io.quarkus.ts.openshift.scaling.ScalingOpenShiftIT  Time elapsed: 104.177 s  <<< ERROR!
java.lang.NoClassDefFoundError: org/apache/http/client/HttpClient
Caused by: java.lang.ClassNotFoundException: org.apache.http.client.HttpClient
```

scaling/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.spec.javax.ws.rs</groupId>
+            <artifactId>jboss-jaxrs-api_2.1_spec</artifactId>
+        </dependency>
```
Ladicek, line 69: Superfluous.

scaling/src/main/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+quarkus.application.name=test-scaling
+quarkus.http.root-path=/api
+quarkus.kubernetes.deployment-target=openshift
```
Ladicek, line 3: All 3 items above are useless IMHO.

Ladicek: Actually I see you depend on the app name, so 2 items above :-)

scaling/src/main/docker/Dockerfile.jvm
```diff
@@ -0,0 +1,34 @@
+####
+# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
+#
+# Before building the docker image run:
+#
+# mvn package
+#
+# Then, build the image with:
+#
+# docker build -f src/main/docker/Dockerfile.jvm -t quarkus/http-jvm .
```
rsvoboda, line 10: `http-jvm` .... copy&paste from http module

Do we need Dockerfiles at all here?

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -0,0 +1,111 @@
...
+
+            assertThat(uniqueIds).hasSize(2);
+        });
+
+        openShiftUtil.scale(DC_NAME, 1);
+
+        assertThat(openShiftUtil.countReadyReplicas(DC_NAME)).isEqualTo(1);
+
+        uniqueIds.clear();
+        for (int i = 0; i < 100; i++) {
```
rsvoboda, line 82: I would add few words why for loop is used instead of `await().atMost(1, TimeUnit.MINUTES).untilAsserted(()`

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -0,0 +1,111 @@
...
+        assertThat(openShiftUtil.countReadyReplicas(DC_NAME)).isEqualTo(1);
+    }
+
+    @Test
+    public void scaleToZero() {
+        openShiftUtil.scale(DC_NAME, 0);
+
+        assertThat(openShiftUtil.countReadyReplicas(DC_NAME)).isEqualTo(0);
+
+        for (int i = 0; i < 100; i++) {
```
rsvoboda, line 103: Why so many requests?

scaling/pom.xml
mjurc: Fixed!

scaling/src/main/resources/application.properties
mjurc: Fixed too

scaling/src/main/docker/Dockerfile.jvm
mjurc: No ass far as I can tell, so I removed them

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
mjurc: I've added workflow to each test method.

Basically the thinking was to gain a slight confidence about the fact that the deployment was scaled down, and it wasn't an intermittent hiccup or such.

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
mjurc: It's an arbitrary amount, if you think it's too much, I guess I can reduce it.

scaling/src/main/docker/Dockerfile.jvm
Ladicek: I have kept the `Dockerfile` in all test applications, because we wanted to test them as well, somehow. (Not sure if that's gonna be possible, but that was the intent.) Indeed they are all copied from `http`, but that doesn't matter, because this is just a comment.

PR #42, opened 2020-05-24
Disabling the Quarkus Hero workshop tests unless test suite is configured to run with authenticated registry.

mjurc:


Ladicek:
Why not use the PostgreSQL 10 image instead? That is available without auth :-)

PR #44, opened 2020-05-27
Rename quarkus-master profile to quarkus-core-only

rsvoboda:
Rename quarkus-master profile to quarkus-core-only

This reflects reality more accurately.


Ladicek:
Thanks!

PR #47, opened 2020-05-29
Created openshift-extension module with tests on Openshift guide

mirostary:
Openshift guide - https://access.redhat.com/documentation/en-us/red_hat_build_of_quarkus/1.3/html-single/deploying_quarkus_applications_on_red_hat_openshift_container_platform/index#proc-using-openshfit-extension_quarkus-openshift

The code is from top of last Petrs branch, only with few changes
- The test downloads quarkus-quickstarts repository and uses getting-started qs
- It sets parameters from the guide (add extension and property file) and deploys QS on Openshift
- Then it changes returning string from the app and deploys again.


mirostary:
Mvnw files moving is a mistake I didn't notice, thank you. And I will make more simple the rest of it.

openshift-extension/src/test/java/io/quarkus/ts/openshift/extension/Command.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.fusesource.jansi.Ansi.ansi;
+
+public class Command {
```
Ladicek, line 12: Why do you copy this class?

openshift-extension/src/test/java/io/quarkus/ts/openshift/extension/ExtensionTest.java
```diff
@@ -0,0 +1,142 @@
...
+            Arrays.asList(
+                    "quarkus.kubernetes-client.trust-certs=true",
+                    "quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7",
+                    "quarkus.openshift.expose=true"
+            ));
+
+    @Test
+    public void checkAddExtension_QuarkusOpenshift() throws IOException, InterruptedException {
+        new Command("git", "clone", QS).runAndWait();
+        if (new File(PATH_APP_PROPERTIES).exists()) {
```
Ladicek, line 53: In general, prefer `Files` and `Path` over `File`.

openshift-extension/src/test/java/io/quarkus/ts/openshift/extension/ExtensionTest.java
```diff
@@ -0,0 +1,142 @@
...
+            new Command("rm", "-f", PATH_APP_PROPERTIES).runAndWait();
```
Ladicek, line 54: `Files.delete`?

openshift-extension/src/test/java/io/quarkus/ts/openshift/extension/OpenShiftClientResource.java
```diff
@@ -0,0 +1,22 @@
+package io.quarkus.ts.openshift.extension;
+
+import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.OpenShiftClient;
+import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
+
+final class OpenShiftClientResource implements CloseableResource {
```
Ladicek, line 7: You don't want this, this is only useful for JUnit 5 extensions. Just create `DefaultOpenShiftClient` directly.

openshift-extension/src/test/java/io/quarkus/ts/openshift/extension/ExtensionTest.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.awaitility.Awaitility.await;
+
+@QuarkusTest
```
Ladicek, line 27: This really shouldn't be `@QuarkusTest`.

Ladicek:
Just added a couple of comments before realizing that, as @rsvoboda says, the entire approach to this test is flawed.

You want a very simple application (similar to the `http` module), which has `quarkus.kubernetes.deploy` in its POM (so that it's always enabled), uses `@OpenShiftTest` and `@ManualApplicationDeployment` (so that the test framework doesn't deploy the app and uses the one that is already deployed) and probably `@CustomizeApplicationUndeployment` to undeploy the app (which amounts to `oc delete -f target/kubernetes/openshift.yml`).

And the name should really be a LOT more descriptive than `openshift-extension`.

Ladicek:
Here's how I believe the test should look like: https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/48

penehyba:
@Ladicek the test was focused on one special row: quarkus:add-extension -Dextensions="openshift"

So we took app that implicitly works (the getting-started example). Then applied the add-extension and check what happens... the test you provided doesn't follow the guide beforementioned.

Ladicek:
Well, it actually does verify the same thing, which is deploying to OpenShift using `quarkus.kubernetes.deploy`.

I do agree that testing the Quarkus Maven plugin features is important, but I think that should be in a dedicated test suite, perhaps directly in the upstream project.

rsvoboda:
Suppressed by https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/48

PR #48, opened 2020-06-01
add test for quarkus.kubernetes.deploy

Ladicek:


deployment-strategies/quarkus/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+
+    <artifactId>deployment-strategies-quarkus</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Quarkus OpenShift TS: Deployment Strategies: Quarkus</name>
+
+    <properties>
+        <!-- this could also be set in application.properties, but this is much closer to what users actually do -->
+        <quarkus.kubernetes.deploy>true</quarkus.kubernetes.deploy>
+        <quarkus.kubernetes-client.trust-certs>true</quarkus.kubernetes-client.trust-certs>
```
rsvoboda, line 22: Why isn't this in `application.properties`?

Ladicek: Because it's only meant for build time. The application doesn't use the Kubernetes client, it's only used during build when the Quarkus Kubernetes extension talks to the OpenShift cluster.

(Yes, I'm aware that the guide says this should be in `application.properties`. It's not wrong, but somewhat misleading IMHO.)

Ladicek: Technically, both these properties could be in `application.properties`, and it would all work. That's actually what I had first :-)

PR #50, opened 2020-06-05
add steps how to debug failing tests

Ladicek:


Ladicek:
Just added one more commit that runs `oc status --suggest` at the end, if tests failed. I think we've talked about that before, but never implemented it.

PR #63, opened 2020-08-10
Updates for test applications with native mode

mjurc:
* Fixing issue with `ScalingOpenShiftIT` where the application instance identifier was determined at compile time
* Adding a `native-image` Quarkus profile to enable build of application with Mandrel builder image and deployment on respective S2I base image
* Adding a readme file section with notes on running this test suite with native image mode

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,27 @@ jobs:
         with:
           name: ci-artifacts
           path: artifacts-jvm${{ matrix.java }}.zip
+  build-released-native:
+    name: Native build - released Quarkus
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 11 ]
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Install JDK {{ matrix.java }}
+        uses: joschi/setup-jdk@v1.0.0
+        with:
+          java-version: openjdk${{ matrix.java }}
+      - name: Build with Maven
+        run: |
+          mvn -V -B clean install -DskipTests -DskipITs -pl 'app-metadata/deployment,app-metadata/runtime,common,http' -Dquarkus.profile=native -Dquarkus.native.container-runtime=docker
   build-master:
     name: JVM build - Quarkus master
     runs-on: ubuntu-latest
@@ -59,3 +80,26 @@ jobs:
         with:
           name: ci-artifacts
           path: artifacts-jvm${{ matrix.java }}.zip
+  build-master-native:
+    name: Native build - Quarkus master
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 11 ]
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Install JDK {{ matrix.java }}
+        uses: joschi/setup-jdk@v1.0.0
+        with:
+          java-version: openjdk${{ matrix.java }}
+      - name: Build Quarkus master
+        run: git clone https://github.com/quarkusio/quarkus.git && cd quarkus && mvn -B clean install -DskipTests -DskipITs -DskipDocs
+      - name: Build with Maven
+        run: |
+          mvn -V -B clean install -Dquarkus-core-only -DskipTests -DskipITs -pl 'app-metadata/deployment,app-metadata/runtime,common,http' -Dquarkus.profile=native -Dquarkus.native.container-runtime=docker
--- a/README.md
+++ b/README.md
@@ -258,6 +258,23 @@ This currently works automatically for the `target/kubernetes/openshift.yml` fil
 
 Note that it is usually a good idea to set `-Dts.image-overrides` to a _full_ path, because Maven changes the current working directory when running tests.
 
+### Native image
+
+The test suite contains a Maven profile configuring required system properties for native image build. The profile is
+activated by specifying the `native` Quarkus profile (`-Dquarkus.profile=native`). This profile does not specify the 
+container runtime for Quarkus native build, so by default, native build will use local GraalVM. The following command 
+will execute the whole test suite using Docker to run containers for native build:
+
+```
+./mvnw clean verify \
+  -Dquarkus.profile=native \ 
+  -Dquarkus.native.container-runtime=docker \
+  -Dts.authenticated-registry
+```
+
+Currently used builder image is `quarkus/ubi-quarkus-mandrel` and the base image for OpenShift deployment is 
+`quarkus/ubi-quarkus-native-binary-s2i`.
+
 ### TODO
 
 There's a lot of possible improvements that haven't been implemented yet.
@@ -277,6 +294,7 @@ The most interesting probably are:
   Currently, this isn't possible, and doesn't make too much sense, because for some actions, we just run `oc`, while for some, we use the Java client.
   We could possibly move towards doing everything through the Java library, but some things are hard (e.g. `oc start-build`).
   If we ever do this, then it becomes easily possible to consisently apply image overrides everywhere.
+- Implement a build workflow with `ubi8/ubi-minimal` instead of using `quarkus/ubi-quarkus-native-binary-s2i`.
 
 It would also be possible to create a Kubernetes variant: `@KubernetesTest`, `kubernetes.yml`, injection of `KubernetesClient`, etc.
 Most of the code could easily be shared.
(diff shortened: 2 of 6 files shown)


Ladicek:
I actually think that instead of modifying `application.properties` for each test, it would be better to add a Maven profile activated by something like `-Dnative`. I have this in my Git stash:

```xml
        <profile>
            <id>native</id>
            <activation>
                <property>
                    <name>native</name>
                </property>
            </activation>

            <properties>
                <quarkus.package.type>native</quarkus.package.type>
            </properties>

            <build>
                <pluginManagement>
                    <plugins>
                        <plugin>
                            <groupId>org.apache.maven.plugins</groupId>
                            <artifactId>maven-failsafe-plugin</artifactId>
                            <executions>
                                <execution>
                                    <id>native-tests</id>
                                    <configuration>
                                        <!-- TODO e.g. use *TestNative.java for native tests? -->
                                        <!-- we need Failsafe for OpenShift tests -->
                                        <systemProperties>
                                            <native.image.path>
                                                ${project.build.directory}/${project.build.finalName}-runner
                                            </native.image.path>
                                        </systemProperties>
                                    </configuration>
                                </execution>
                            </executions>
                        </plugin>
                    </plugins>
                </pluginManagement>
            </build>
        </profile>
```

This sets `quarkus.package.type`, it could also set `quarkus.native.builder-image` and perhaps even `quarkus.s2i.base-jvm-image` . Native-specific configuration would still have to be solved somehow, but that would only be a small subset of the tests (I can only think of the ConfigMap test ATM). Using a profile with a well-known name is a good idea.

This would also allow running native tests (Quarkus `@NativeTest`).

Ladicek:
Or, instead of setting `quarkus.s2i.base-jvm-image` in the POM (which I agree is a questionable idea and would bring troubles, e.g. in multi-arch testing), perhaps we could leverage the image overrides feature.

mjurc:
@Ladicek I've used the image overrides for testing, for some reason the Quarkus profile seemed nicer to me - I mean, it's persisted right in the repo and ready to go.

On the other hand, if we wanted to use overrides, I guess they could be activated by a property in activated (Maven) profile too.

I'm not really leaning towards any of those solutions, so I just went the way the Quarkus docs say to do it

Ladicek:
So when you want to change the builder image (which you will have to do often, currently you have `quay.io/quarkus/ubi-quarkus-mandrel:20.1.0.1.Alpha2-java11`), you will edit all the `application.properties` files?

The Maven profile approach also does exactly what the Quarkus docs say, the properties are just centralized on a single place.

mjurc:
Yeah, I agree actually, I'm gonna go the way of Maven profile

README.md
```diff
@@ -233,6 +233,23 @@ This currently works automatically for the `target/kubernetes/openshift.yml` fil
 
 Note that it is usually a good idea to set `-Dts.image-overrides` to a _full_ path, because Maven changes the current working directory when running tests.
 
+### Native image
+
+The test suite contains a Maven profile configuring required system properties for native image build. The profile is
+activated by specifying the `native` Quarkus profile (`-Dquarkus.profile=native`). This profile does not specify the 
+container runtime for Quarkus builder image, so by default, native build will use local GraalVM. The following command
```
Ladicek, line 240: "container runtime for Quarkus builder image" -- the term "builder image" is overloaded, I'd use "container runtime for Quarkus native build".

README.md
```diff
@@ -233,6 +233,23 @@ This currently works automatically for the `target/kubernetes/openshift.yml` fil
 
 Note that it is usually a good idea to set `-Dts.image-overrides` to a _full_ path, because Maven changes the current working directory when running tests.
 
+### Native image
+
+The test suite contains a Maven profile configuring required system properties for native image build. The profile is
+activated by specifying the `native` Quarkus profile (`-Dquarkus.profile=native`). This profile does not specify the 
+container runtime for Quarkus builder image, so by default, native build will use local GraalVM. The following command 
+that will execute the whole test suite using Docker to run containers for native build:
```
Ladicek, line 241: Remove "that" at the beginning of the line.

pom.xml
```diff
@@ -242,5 +242,23 @@
...
+            <id>native</id>
+            <activation>
+                <property>
+                    <name>quarkus.profile</name>
+                    <value>native</value>
+                </property>
+            </activation>
+
+            <properties>
+                <quarkus.profile>native</quarkus.profile>
```
Ladicek, line 256: Remove this line, to avoid impression that it might have an effect.

Ladicek:
Added a few comments, otherwise LGTM.

README.md
```diff
@@ -233,6 +233,23 @@ This currently works automatically for the `target/kubernetes/openshift.yml` fil
...
+  -Dquarkus.native.container-runtime=docker \
+  -Dts.authenticated-registry
+```
+
+Currently used builder image is `quarkus/ubi-quarkus-mandrel` and the base image for OpenShift deployment is 
+`quarkus/ubi-quarkus-native-binary-s2i`.
+
 ### TODO
 
 There's a lot of possible improvements that haven't been implemented yet.
```
rsvoboda, line 255: Add item to avoid usage of `quarkus/ubi-quarkus-native-binary-s2i`, in the end ubi8-minimal should be used for the product

rsvoboda:
https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/master/.github/workflows/ci.yml should be updated to compile at least one module into native - e.g. http

README.md
Ladicek: `quarkus/ubi-quarkus-native-binary-s2i` is actually `ubi8-minimal` + a few packages + the S2I scripts. I'm not sure if it's meant to stay like this, but if it's a TODO, then primarily somewhere else (and only secondarily in this test suite).

rsvoboda: It's todo for this TS, ubi-quarkus-native-binary-s2i is for community use, not supported for the product 😎
I'm fine with using it for now because of the reasons you mentioned (being actually ubi8-minimal + few things on top of it). Just want to make sure this item is tracked.

Ladicek:
One thing I just realized: the `quay.io/quarkus/ubi-quarkus-native-binary-s2i:1.0` S2I image we currently use installs a few packages on top of `ubi8-minimal`. I have no idea if this is an artifact of sharing the same Cekit module with other Quarkus images, or if some of these packages are actually required for running GraalVM-produced binaries. We need to verify that.

pom.xml
rsvoboda: @mjurc you forgot about this?

configmap/src/main/resources/application.properties
```diff
@@ -4,3 +4,5 @@ quarkus.openshift.mounts.app-config.path=/deployments/config
 quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
 
 %test.hello.message=Hello, %s!
+
+%native.quarkus.openshift.mounts.app-config.path=/home/quarkus/config
```
rsvoboda, line 8: Add here explanation why this is needed / pointer to GH issue discussion

.github/workflows/ci.yml
```diff
@@ -30,6 +30,27 @@ jobs:
...
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Install JDK {{ matrix.java }}
+        uses: joschi/setup-jdk@v1.0.0
+        with:
+          java-version: openjdk${{ matrix.java }}
+      - name: Build with Maven
```
rsvoboda, line 51: I don't think it's necessary to have 2 new builds, we could have just native step added to the existing builds.

What's your motivation to have separate native builds? Is it to have separated native vs. jvm ++ more concurrent builds / earlier feedback ...
The downside of this approach is that more instances are needed (not big deal for us) ++ Quarkus master is build twice.

An example without adding new builds:
```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index eca0e1a..5c49744 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   - pull_request
 jobs:
   build-released:
-    name: JVM build - released Quarkus
+    name: Build - released Quarkus
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -23,6 +23,9 @@ jobs:
           java-version: openjdk${{ matrix.java }}
       - name: Build with Maven
         run: mvn -V -B clean test
+      - name: Native Build
+        run: |
+          mvn -V -B clean package -DskipTests -DskipITs -pl 'app-metadata/deployment,app-metadata/runtime,common,http' -Dquarkus.profile=native -Dquarkus.native.container-runtime=docker
       - name: Zip Artifacts
         run: |
           zip -R artifacts-jvm${{ matrix.java }}.zip 'surefire-reports/*'
@@ -31,7 +34,7 @@ jobs:
           name: ci-artifacts
           path: artifacts-jvm${{ matrix.java }}.zip
   build-master:
-    name: JVM build - Quarkus master
+    name: Build - Quarkus master
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -52,6 +55,9 @@ jobs:
         run: git clone https://github.com/quarkusio/quarkus.git && cd quarkus && mvn -B clean install -DskipTests -DskipITs -DskipDocs
       - name: Build with Maven
         run: mvn -V -B clean test -Dquarkus-core-only
+      - name: Native Build
+        run: |
+          mvn -V -B clean package -Dquarkus-core-only -DskipTests -DskipITs -pl 'app-metadata/deployment,app-metadata/runtime,common,http' -Dquarkus.profile=native -Dquarkus.native.container-runtime=docker
       - name: Zip Artifacts
         run: |
           zip -R artifacts-jvm${{ matrix.java }}.zip 'surefire-reports/*'
```

rsvoboda: On the second thought, I think we can proceed with this approach and optimize for resources later / if needed.
This gives us better view in PRs

PR #65, opened 2020-08-11
Default to product coordinates

rsvoboda:
Default to product coordinates, for community coordinates `quarkus-core-only` profile can be used

`quarkus-universe-bom` is only available under `com.redhat.quarkus` groupId for RHBQ

Motivation: We should have consistent way to run the TS on product side (e.g. z-series team, interop testing)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Build Quarkus master
         run: git clone https://github.com/quarkusio/quarkus.git && cd quarkus && mvn -B clean install -DskipTests -DskipITs -DskipDocs
       - name: Build with Maven
-        run: mvn -V -B clean test
+        run: mvn -V -B clean test -Dquarkus-core-only
       - name: Zip Artifacts
         run: |
           zip -R artifacts-jvm${{ matrix.java }}.zip 'surefire-reports/*'
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ All tests will run in the project you're logged into, so it should be empty.
 If there are resources deployed in the project, you should not expect they will survive.
 
 Running the tests amounts to standard `mvn clean verify`.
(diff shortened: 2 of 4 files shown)


rsvoboda:
"awesome", [INFO] Quarkus OpenShift TS: Messaging: AMQP with Reactive Messaging FAILURE [ 24.469 s]

Not related to my changes but kills the joy :/

Ladicek:
This is wrong. At the very least, the POM doesn't declare MRRC, so running plain old `mvn clean verify` will most likely fail. (If you have MRRC in your `~/.m2/settings.xml`, you're a tiny minority.)

Second, related to the branching strategy: the `master` branch is for latest upstream/downstream development, yet it forces RHBQ 1.3.4. I have several tests written locally that won't even compile with this setting. At this point, I'm starting to think if branches shouldn't default to the product version, while `master` would default to the latest upstream. WDYT?

rsvoboda:
There are two views on this ... yours and ours :) We use master for product testing (1.3.4 is there because it's the latest public availabe) and we branch once the product is released. This PR make the command used on Jenkins unified for 1.3 and master, this command is used by other teams too (PE for example).

For community bits the default command is `mvn clean verify -Dquarkus-core-only`

Ladicek:
Then the branching strategy you wrote down (copied from Thorntail :-) ) is wrong. And there's no way I can submit my new tests to `master`. We need to figure this out, because Quarkus moves fast and breaking changes are inevitable at this point.

I really think the 1.3 branch must be used for 1.3 testing, otherwise `master` can't contain tests that require 1.4+ and don't compile on 1.3.

rsvoboda:
The trouble is that we moved to `quarkus-universe-bom` and it has different groupId on product side vs. project side.
For `quarkus-bom` there are productized both `io.quarkus` and `com.redhat.quarkus` groupId

rsvoboda:
You can add new test - https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/master/.github/workflows/ci.yml#L27

Issue #66, opened 2020-08-11
AMQP with Reactive Messaging fails when running against Quarkus master / 1.7.0.Final


rsvoboda:
AMQP with Reactive Messaging fails when running against Quarkus master / 1.7.0.Final

Spotted in https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/65
[INFO] Quarkus OpenShift TS: Messaging: AMQP with Reactive Messaging FAILURE [ 24.469 s]

Reproducer: mvn clean test -Dquarkus-core-only -Dquarkus.http.host=127.0.0.1

rsvoboda:
Test passes with 1.7.0.Final when I explicitlly downgrade smallrye-reactive-messaging-amqp from 2.2.1 to 2.1.1

```xml
        <dependency>
            <groupId>io.smallrye.reactive</groupId>
            <artifactId>smallrye-reactive-messaging-amqp</artifactId>
            <version>2.1.1</version>
        </dependency>
```

rsvoboda:
Looking also at https://github.com/quarkusio/quarkus-quickstarts/commits/development/amqp-quickstart

rsvoboda:
https://github.com/quarkusio/quarkus/commit/0dd61af3e6a7e6d92f121d53ebb85b429416ffce#diff-137f7428dbac3e581b5a738242179b31 introduces the change in behavior / bugfix

PR #68, opened 2020-08-11
Fix the durability when sending messages to a non-durable AMQP address, enforced with reactive-messaging 2.2

rsvoboda:
Fix the durability when sending messages to a non-durable AMQP address, enforced with reactive-messaging 2.2

Fixes https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/66

Introduced in https://github.com/quarkusio/quarkus/commit/0dd61af3e6a7e6d92f121d53ebb85b429416ffce

### DIFF
--- a/messaging/amqp-reactive/src/main/resources/application.properties
+++ b/messaging/amqp-reactive/src/main/resources/application.properties
@@ -11,7 +11,6 @@ amqp-port=5672
 # Configure the AMQP connector to write to the `prices` address
 mp.messaging.outgoing.generated-price.connector=smallrye-amqp
 mp.messaging.outgoing.generated-price.address=prices
-mp.messaging.outgoing.generated-price.durable=true
 
 # Configure the AMQP connector to read from the `prices` queue
 mp.messaging.incoming.prices.connector=smallrye-amqp


messaging/amqp-reactive/src/main/resources/application.properties
```diff
@@ -11,7 +11,6 @@ amqp-port=5672
 # Configure the AMQP connector to write to the `prices` address
 mp.messaging.outgoing.generated-price.connector=smallrye-amqp
 mp.messaging.outgoing.generated-price.address=prices
-mp.messaging.outgoing.generated-price.durable=true
```
Ladicek, line 14: What I find really weird, at least from my perspective (I don't know almost anything about messaging) is that you delete the `durable` setting for `outgoing`, but not for `incoming`, even if they address the same queue.

Ladicek: Hmm, or they don't use the same queue? The comments suggest they do, but `incoming` doesn't have the `address` setting at all. Oh well.

rsvoboda: I'm asking for some input on https://github.com/quarkusio/quarkus/commit/0dd61af3e6a7e6d92f121d53ebb85b429416ffce feel free to comment there too

Same change is in quarkus-quickstarts ....

Ladicek: Thanks.

PR #71, opened 2020-08-11
new tests

Ladicek:


Ladicek:
This conflicts with the PR for native binaries. The new ConfigMap (and most likely also Secret) tests don't work with native (same reason as in the PR for native binaries), so I'm fine with merging that PR first. I'll rebase this one then, and fix the native stuff.

Submitting just for preliminary review.

Ladicek:
Fixed paths to parent POMs.

Ladicek:
Improved HTTPS tests to expect more general `SSLException`, see the commit message for more details.

Ladicek:
Now _this_ failure is weird...

Ladicek:
Aha! I still use Java 8 by default locally, and there it passes. It fails with Java 11. Interesting.

Ladicek:
I need to work on something else, so I'll postpone the investigation for a few days. I wanted to merge this after the native stuff anyway.

rsvoboda:
@Ladicek can you split this PR into more PRs? At least move two-way TLS into separate one.

Ladicek:
I can do that, but also this PR has the missing `@RegisterForReflection` annotations and `%native.quarkus.openshift.whatever` properties required for native. So I'd prefer getting the native PR merged first -- which will hopefully take enough time for me to investigate the problem here :-)

rsvoboda:
OK :)

Ladicek:
OK, I know what's wrong. With Java 11, the HTTPS client automatically negotiates TLSv1.3 with the server, while on Java 8, it negotiates TLSv1.2. If I enforce TLSv1.2 on Java 11, the tests pass just fine. I wasn't able to enforce TLSv1.3 with my Java 8, even though I have OpenJDK 8u265, which should support it.

To be sure, I also tried with Java 12, 13 and 14, and the test behaves identically to Java 11. (That is, fails by default, but passes when TLSv1.2 is enforced.)

I don't know why there's such a behavioral difference.

Considering Quarkus attempts to get rid of Java 8 and establish Java 11 as the base Java version (which is essentially a done deal for native, albeit not for JVM), I don't think it makes sense to enforce TLSv1.2. (And that is even though the HTTPS tests are only `@QuarkusTest` and don't deploy anything to OpenShift, so they don't really use native binaries.) Instead, I'll just make the test decide based on current Java version.

Ladicek:
Quarkus master fails to build, because `Connection timed out (Read failed)`.

Ladicek:
Otherwise I think this is ready.

rsvoboda:
Tests passed for JVM mode

I see failures in NATIVE mode - https://quarkus-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/openshift-test-suite-native-ci/1/testReport/

io.quarkus.ts.openshift.config.system.api.server.ConfigSecretApiServerOpenShiftIT
io.quarkus.ts.openshift.config.secret.file.system.ConfigSecretFileSystemOpenShiftIT
io.quarkus.ts.openshift.configmap.api.server.ConfigMapApiServerOpenShiftIT

Ladicek:
That's not funny, I did all the changes @mjurc did for ConfigMap test! I'll look into it, thanks.

configmap/api-server/src/main/resources/application.properties
```diff
@@ -0,0 +1,10 @@
+quarkus.openshift.expose=true
+quarkus.kubernetes-config.enabled=true
```
rsvoboda, line 2: So this is the main difference between fs vs. api approach?

configmap.yaml files look the same.

security/https-2way-authz/src/main/java/io/quarkus/ts/openshift/security/https/twoway/authz/CertificateRolesAugmentor.java
```diff
@@ -0,0 +1,54 @@
...
+import io.smallrye.mutiny.Uni;
+
+import javax.inject.Singleton;
+
+import java.security.cert.X509Certificate;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Quarkus doesn't have a declarative/configuration way to assign roles when using certificate authentication.
```
rsvoboda, line 17: Should be there an feature request for this?

security/https-2way-authz/src/test/java/io/quarkus/ts/openshift/security/https/twoway/authz/SecurityHttps2wayAuthzTest.java
```diff
@@ -0,0 +1,158 @@
...
+                .loadTrustMaterial(new File("target/client-truststore.pkcs12"), CLIENT_PASSWORD)
+                .build();
+        try (CloseableHttpClient httpClient = HttpClients.custom()
+                .setSSLContext(sslContext)
+                .setSSLHostnameVerifier(new DefaultHostnameVerifier())
+                .build()) {
+
+            Executor executor = Executor.newInstance(httpClient);
+
+            {
```
rsvoboda, line 55: Why is this `{ ... }` approach needed?

configmap/api-server/src/main/resources/application.properties
Ladicek: Yes. Instead of mounting the ConfigMap into the filesystem, we use the `quarkus-kubernetes-config` extension, which creates another `ConfigSource`, which reads the `application.properties` file directly from the Kubernetes API server.

security/https-2way-authz/src/main/java/io/quarkus/ts/openshift/security/https/twoway/authz/CertificateRolesAugmentor.java
Ladicek: Perhaps. Not sure.

security/https-2way-authz/src/test/java/io/quarkus/ts/openshift/security/https/twoway/authz/SecurityHttps2wayAuthzTest.java
Ladicek: It's not needed, but I believe it improves readability. There are 2 independent requests, so I have 2 independent blocks of code. If I didn't, I'd have to either reuse the `response` variable, or have a different name for the 2nd one, and I find both approaches more error-prone than this.

security/https-2way-authz/src/main/java/io/quarkus/ts/openshift/security/https/twoway/authz/CertificateRolesAugmentor.java
rsvoboda: I'm not sure either :) Code in this file looks quite specific to the structure of the keystore, especially `extractRoles` method.

security/https-2way-authz/src/test/java/io/quarkus/ts/openshift/security/https/twoway/authz/SecurityHttps2wayAuthzTest.java
rsvoboda: ok, thanks for explaining

security/https-2way-authz/src/main/java/io/quarkus/ts/openshift/security/https/twoway/authz/CertificateRolesAugmentor.java
rsvoboda: For this case I can imagine declarative mapping via application.properties like this
```
quarkus.http.ssl.certificate.role-mapping.cn.client=user
quarkus.http.ssl.certificate.role-mapping.cn.guest-client=guest
```
This would cover just some (simple) use-case for role mapping.
wdyt?

Ladicek: WildFly supports this, both with legacy security and Elytron. In both cases, the  configuration is quite crazy. See:

- legacy security:
    - https://github.com/rhoar-qe/thorntail-test-suite/blob/master/protocols/https-2way-authz/src/main/resources/project-defaults.yml
    - https://github.com/rhoar-qe/thorntail-test-suite/blob/master/protocols/https-2way-authz/src/main/resources/roles.properties
- Elytron:
    - https://github.com/rhoar-qe/thorntail-test-suite/blob/master/protocols/https-2way-authz-elytron/src/main/resources/project-defaults.yml
    - https://github.com/rhoar-qe/thorntail-test-suite/blob/master/protocols/https-2way-authz-elytron/src/main/resources/roles.properties

Hence, I created this class in a generic way. The only thing you need to change is the `extractRoles` method.

Ladicek: I filed https://github.com/quarkusio/quarkus/issues/11434

rsvoboda: Thanks!

Ladicek:
I've pushed a fix. Note that `ConfigSecretFileSystemOpenShiftIT` failed because

```
route/config-secret-file-system was not accepted by router "default": apps.ocp45.dynamic.quarkushost name validation errors: spec.host: Invalid value: "config-secret-file-system-project-openshift-test-suite-native-ci.apps.ocp45.dynamic.quarkus": must be no more than 63 characters (InvalidHost)
```

You really shouldn't use Jenkins job name as OpenShift project name, it's too long. A random string or some hash of the job name, trimmed down to cca 10 chars, should be perfectly fine.

rsvoboda:
ok, I will trim the name like this:
```bash
JOB_NAME=openshift-test-suite-native-ci
JOB_MD5=`echo $JOB_NAME | md5sum`
OCP_PROJECT=${JOB_MD5:0:10}
echo $OCP_PROJECT
```

rsvoboda:
No failures in https://quarkus-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/openshift-test-suite-native-ci/4/
Congrats 👍

Issue #70, opened 2020-08-11
LogFactory issue for external-applications


rsvoboda:
`mvn clean verify -Dquarkus-core-only -Dversion.quarkus=1.7.0.Final -Dquarkus.http.host=127.0.0.1 -pl app-metadata/deployment,app-metadata/runtime,common,external-applications/todo-demo-app`

gives
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 135.115 s <<< FAILURE! - in io.quarkus.ts.openshift.todo.demo.app.TodoDemoAppOpenShiftIT
[ERROR] io.quarkus.ts.openshift.todo.demo.app.TodoDemoAppOpenShiftIT Time elapsed: 135.114 s <<< ERROR!
java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory
Caused by: java.lang.ClassNotFoundException: org.apache.commons.logging.LogFactory

[INFO]
[INFO] Results:
[INFO]
[ERROR] Errors:
[ERROR] TodoDemoAppOpenShiftIT » NoClassDefFound org/apache/commons/logging/LogFactory
[INFO]
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
```

rsvoboda:
The issue is not in the application itself, checked with modified https://github.com/rsvoboda/todo-demo-app

org.apache.commons.logging.LogFactory is present in `commons-logging:commons-logging:1.2`
There are plenty of exlusions of `commons-logging:commons-logging` in https://github.com/quarkusio/quarkus/blob/master/bom/application/pom.xml

Might be the case that some dependency somehow requires commons-logging to be present.
There are some issues related to LogFactory in  https://github.com/quarkusio/quarkus/issues?q=is%3Aissue+is%3Aopen+LogFactory But nothing really close to our case.

rsvoboda:
Played with various exclusions in `common` module, but what worked was to explicitly add `commons-logging:commons-logging:1.2` dependency.

rsvoboda:
One more thing - I'm not really able to locate where the need for commons-logging comes from, I wasn't able to get stacktrace.

Ladicek:
Funny, I took your fork of the example app, reverted the last commit, and still `mvn clean package` passes just fine.

rsvoboda:
You need to run `verify`, IT test execution fails

[...]

PR #72, opened 2020-08-11
Add commons-logging dependency to avoid NoClassDefFound org/apache/commons/logging/LogFactory

rsvoboda:
Add commons-logging dependency to avoid NoClassDefFound org/apache/commons/logging/LogFactory

Fixes https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/70, details are also described in https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/70

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -46,6 +46,10 @@
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>commons-logging-jboss-logging</artifactId>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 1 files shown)


Ladicek:
Should we use `org.jboss.logging:commons-logging-jboss-logging` instead? That should be managed by the Quarkus BOM.

Ladicek:
... actually Quarkus should use `org.jboss.logging:commons-logging-jboss-logging` always. Perhaps there's something in Quarkus that doesn't propagate it transitively even if it should...?

rsvoboda:
oki, I will use it. As I wrote in #70 I couldn't get stacktrace to understand where is it coming from :/

rsvoboda:
Quarkus master fails to build, not related to my changes ;)

common/pom.xml
```diff
@@ -46,6 +46,10 @@
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
         </dependency>
+        <dependency>
```
Ladicek, line 49:
```suggestion
        <!-- RestAssured depends on Apache HTTP Client, which depends on Commons Logging -->
        <dependency>
```

PR #73, opened 2020-08-11
quarkus-builder is in BOM starting with Quarkus 1.6

rsvoboda:
quarkus-builder is in BOM starting with Quarkus 1.6


rsvoboda:
Checks are green, merging

PR #76, opened 2020-08-13
Execute GH Action only with PR, no need to re-run it with merge

rsvoboda:
Execute GH Action only with PR, no need to re-run it with merge


rsvoboda:
Timeout on one GH action, ignoring as it's env related.

PR #80, opened 2020-08-27
OpenShift Serverless support in test framework, first test for OpenShift Serverless

rsvoboda:
OpenShift Serverless support in test framework, first test for OpenShift Serverless

See changes in README.md

Went ahead with dedicated profile because `@OnlyIfConfigured("ts.serverless")` approach wouldn't prevent execution of the S2I build during the `package` phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,10 +227,10 @@ In such case, the `reason` attribute of the annotation should point to correspon
 
 If the test class is annotated `@ManualApplicationDeployment`, the `target/kubernetes/openshift.yml` file is ignored and the test application is _not_ deployed automatically.
 Instead, you should use `@AdditionalResources`, `@CustomizeApplicationDeployment` and `@CustomizeApplicationUndeployment` to deploy the application manually.
-The `@ManualApplicationDeployment` annotation then provides the necessary info about the application (such as the name or a known endpoint).
 
 This can be used to write tests that excersise an external application.
 You have full control over the application deployment and undeployment process, but the rest of the test can be written as if the application was part of the test suite.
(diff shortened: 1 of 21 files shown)


Ladicek:
This looks pretty good, but reveals a design mistake I made when introducing `@ManualApplicationDeployment`. This annotation conflates 2 things: intent to deploy the application manually, and source of application metadata when the `app-metadata` Quarkus extension is not present. As you just found, sometimes we want to deploy the application manually, yet we don't need to provide app metadata explicitly, because the application source is in the test suite and has the `app-metadata` extension.

This really should be 2 different annotations: `@ManualApplicationDeployment` would just state "I'll deploy the app manually" and would have no attributes, and something like `@ExplicitAppMetadata` would have all the attributes and would only be used if `app-metadata` is not present.

If you wish, I can do the refactoring.

Ladicek:
Here it is: https://github.com/Ladicek/quarkus-openshift-test-suite/commits/serverless

Ladicek:
I can force-push to your branch (to change this PR), if you agree.

rsvoboda:
@Ladicek agree, please proceed

Ladicek:
Done :-)

rsvoboda:
+1 for the changes you did

Ladicek:
Thanks :-) Not sure why the messaging test failed, but perhaps we should invoke Maven with `-fae` in the CI actions.

rsvoboda:
messaging test fails from time to time on GH actions :( Have AI to look into it.

PR #98, opened 2020-10-09
prerequisites for updating to Quarkus 1.9

Ladicek:
This mainly means moving from `container-image-s2i` to
`container-image-openshift`. The `openshift` extension,
which used to bring `container-image-s2i`, now brings
`container-image-openshift`, so wherever we used to use
`container-image-s2i` explicitly (= the `http` test),
we now use `container-image-openshift` instead.

That mainly means one important change, which is largely
invisible to people who exclusively use Quarkus tooling:
the expectation now is that the application files are
present in `/deployments/target` in the container image.
Previously, they were in `/deployments`.

In addition to that, this commit also deals with one
change in Fabric8 Kubernetes Client: they have split
their `Readiness` class into two (Kubernetes-specific
and OpenShift-specific). As a response, this commit
adds `ReadinessUtil` which unifies that.

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
@@ -149,7 +149,14 @@ final class OpenShiftTestExtension implements BeforeAllCallback, AfterAllCallbac
 
             awaitImageStreams(context, openshiftResources);
 
-            new Command("oc", "start-build", getAppMetadata(context).appName, "--from-dir=target", "--follow").runAndWait();
+            // when generating Kubernetes resources, Quarkus expects that all application files
+            // will reside in `/deployments/target`, but if we did just `oc start-build my-app --from-dir=target`,
+            // the files would end up in `/deployments`
+            // using `tar` works around that nicely, because the tarball created with this command will have
+            // a single root directory named `target`, which the S2I builder image will unpack to `/deployments`
+            new Command("tar", "czf", "app.tar.gz", "target").runAndWait();
+            new Command("oc", "start-build", getAppMetadata(context).appName, "--from-archive=app.tar.gz", "--follow").runAndWait();
+            new Command("rm", "app.tar.gz").runAndWait();
         }
 
         setUpRestAssured(context);
--- a/common/src/main/java/io/quarkus/ts/openshift/common/util/ReadinessUtil.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/util/ReadinessUtil.java
@@ -1,0 +1,29 @@
+package io.quarkus.ts.openshift.common.util;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.client.internal.readiness.Readiness;
+import io.fabric8.openshift.api.model.DeploymentConfig;
+import io.fabric8.openshift.client.internal.readiness.OpenShiftReadiness;
+
+/**
+ * Unifies Fabric8 Kubernetes Client's {@link Readiness} and {@link OpenShiftReadiness}.
+ * No other class should refer to them directly; using {@code ReadinessUtil} is preferred.
+ */
+public class ReadinessUtil {
+    public static boolean isReadinessApplicable(Class<? extends HasMetadata> resourceClass) {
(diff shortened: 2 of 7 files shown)


common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
```diff
@@ -149,7 +149,14 @@ private void doBeforeAll(ExtensionContext context) throws Exception {
 
             awaitImageStreams(context, openshiftResources);
 
-            new Command("oc", "start-build", getAppMetadata(context).appName, "--from-dir=target", "--follow").runAndWait();
+            // when generating Kubernetes resources, Quarkus expects that all application files
+            // will reside in `/deployments/target`, but if we did just `oc start-build my-app --from-dir=target`,
+            // the files would end up in `/deployments`
+            // using `tar` works around that nicely, because the tarball created with this command will have
+            // a single root directory named `target`, which the S2I builder image will unpack to `/deployments`
+            new Command("tar", "czf", "app.tar.gz", "target").runAndWait();
```
rsvoboda, line 157: Can we assume `tar` being available on all the systems or there should be explicit note in README?

common/src/main/java/io/quarkus/ts/openshift/common/util/ReadinessUtil.java
```diff
@@ -0,0 +1,29 @@
...
+import io.fabric8.openshift.api.model.DeploymentConfig;
+import io.fabric8.openshift.client.internal.readiness.OpenShiftReadiness;
+
+/**
+ * Unifies Fabric8 Kubernetes Client's {@link Readiness} and {@link OpenShiftReadiness}.
+ * No other class should refer to them directly; using {@code ReadinessUtil} is preferred.
+ */
+public class ReadinessUtil {
+    public static boolean isReadinessApplicable(Class<? extends HasMetadata> resourceClass) {
+        // unfortunately `OpenShiftReadiness` doesn't have a `isReadinessApplicable` method
```
rsvoboda, line 15: design flaw ?

rsvoboda (approved):
approving, only minor comments / questions

common/src/main/java/io/quarkus/ts/openshift/common/util/ReadinessUtil.java
Ladicek: More like an implementation omission. I wanted to submit a PR and then forgot, thanks for reminder :-)

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Ladicek: Hmm. Good point. Or I can bring in a Java implementation of `tar`... which I really wouldn't like to do :-) I mean, `tar` is everywhere :-) But adding a "prerequisites" section to the readme is a good idea.

rsvoboda:
CI will fail anyway on datacource config issue :/

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
rsvoboda: +1 for "prerequisites" section to the readme
no need to pull in java-tar

common/src/main/java/io/quarkus/ts/openshift/common/util/ReadinessUtil.java
rsvoboda: Once that in and released, this util can be simplified .. I guess

Ladicek:
I just converted this to _draft_ to highlight this can't be merged yet. In fact this should go in _together_ with the update to 1.9.

rsvoboda:
I was thinking about pulling it in and bumping to quarkus version 1.9.0.CR1

We would need to use `-Dquarkus-core-only` next 2 weeks

Ladicek:
We also need this: https://github.com/quarkusio/quarkus/pull/12631 Hopefully this will make it into 1.9.0.Final.

Ladicek:
Added prerequisites to README.

Ladicek:
FYI added the missing method in `OpenShiftReadiness` in Fabric8 Kubernetes Client here: https://github.com/fabric8io/kubernetes-client/pull/2558

rsvoboda:
I see failure with QuarkusDeploymentStrategyOpenShiftServerlessIT in the CI run.
But I think we can merge this and look into it in separate PR.

Issue #93, opened 2020-10-06
Investigate and fix instabilities in messaging tests


mjurc:
As of Quarkus 1.7, the messaging tests have been consistently failing in both `surefire` and `failsafe` stage.

The testing deployments and the test assertions need to be reviewed.

Sgitario:
I've been trying to investigate the instabilities in the messaging tests and I found a couple of issues:

- In amqp-reactive without running on Openshift: the AbstractAMQPTest is expecting a list of prices [10 20 ... 100], because the producer sends a price every 1 second. The problem is when the producer emits a new tick after the 10th price before the test checks the result, it will produce another 10 value and the test will not work as the list of prices will be [10 20 ... 100 10]. 

This can be easily fixed by either limiting the producer to emit only 10 values or change the expectation to check whether the result contains 10 20 ... 100 (not equal).

- In amqp-reactive again, but running with Openshift, it seems that the application randomly hangs to connect with the AMQ running in the namespace. I could see this exception:

```
WARNING [io.ver.cor.imp.BlockedThreadChecker] (vertx-blocked-thread-checker) Thread Thread[vert.x-eventloop-thread-0,5,main]=Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 3764 ms, time limit is 2000 ms: io.vertx.core.VertxException: Thread blocked
	at java.base@11.0.8/java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
	at java.base@11.0.8/java.net.InetAddress$PlatformNameService.lookupAllHostAddr(InetAddress.java:929)
	at java.base@11.0.8/java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1515)
	at java.base@11.0.8/java.net.InetAddress$NameServiceAddresses.get(InetAddress.java:848)
```

I need further investigation about how to workaround this issue.

My question now is whether you experienced other or the same issues (as I only spotted flaky tests in the same project "amqp-reactive"). If so, can we see the traces somewhere?

rsvoboda:
@mjurc will ping you and provide link to the job where you can experiment with custom branches etc.
We do not keep history of these fails, only last 10 runs.

Sgitario:
> I've been trying to investigate the instabilities in the messaging tests and I found a couple of issues:
> 
[...]

PR #106, opened 2020-11-02
Enable messaging tests and fix amqp reactive test case

Sgitario:
This PR will enable the messaging tests and fix the amqp-reactive module. Also:
- Fix await route util to ignore exceptions (sometimes it returns 503 Gateway Timeout because the route is not ready yet).
- Avoid to request for resources too often (it can cause too many requests exception -> this happens when the cluster is too busy)
- Create a messaging parent to include all the messaging modules.

See dependant issue: https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/93

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
@@ -36,7 +36,7 @@ public final class AwaitUtil {
     public void awaitAppRoute() {
         System.out.println(ansi().a("waiting for route ").fgYellow().a(metadata.appName).reset()
                 .a(" to start responding at ").fgYellow().a(metadata.knownEndpoint).reset());
-        await().atMost(5, TimeUnit.MINUTES).untilAsserted(() -> {
+        await().ignoreExceptions().atMost(5, TimeUnit.MINUTES).untilAsserted(() -> {
             given()
                     // known endpoint is already httpRoot-adjusted
                     .basePath("/")
@@ -53,7 +53,9 @@ public final class AwaitUtil {
                 .forEach(it -> {
                     System.out.println(ansi().a("waiting for ").a(readableKind(it.getKind())).a(" ")
                             .fgYellow().a(it.getMetadata().getName()).reset().a(" to become ready"));
-                    await().atMost(5, TimeUnit.MINUTES).until(() -> {
+                    await().pollInterval(1, TimeUnit.SECONDS)
+                           .atMost(5, TimeUnit.MINUTES)
+                           .until(() -> {
                         HasMetadata current = oc.resource(it).fromServer().get();
                         if (current == null) {
                             ResourceHandler<HasMetadata, ?> handler = Handlers.get(it.getKind(), it.getApiVersion());
--- a/messaging/amqp-reactive/src/main/resources/application.properties
+++ b/messaging/amqp-reactive/src/main/resources/application.properties
@@ -1,9 +1,6 @@
 quarkus.openshift.expose=true
 quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
 
-# Configures the AMQP broker credentials.
-amqp-username=quarkus
-amqp-password=quarkus
 %test.amqp-host=localhost
 amqp-host=amq-broker-amqp
 amqp-port=5672
--- a/messaging/amqp-reactive/src/test/java/io/quarkus/ts/openshift/messaging/amqp/AbstractAMQPTest.java
+++ b/messaging/amqp-reactive/src/test/java/io/quarkus/ts/openshift/messaging/amqp/AbstractAMQPTest.java
@@ -9,11 +9,17 @@ import java.util.concurrent.TimeUnit;
 
 import static io.restassured.RestAssured.when;
 import static org.awaitility.Awaitility.await;
-import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.containsString;
 
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 public abstract class AbstractAMQPTest {
 
+    /**
+     * The producer sends a price every 1 sec {@link io.quarkus.ts.openshift.messaging.amqp.PriceProducer#generate()}.
+     * Eventually, the consumer will get up to 10 prices (from 10 to 100) but it might receive more
+     * {@link io.quarkus.ts.openshift.messaging.amqp.PriceConsumer#process()}.
+     *
+     */
     @Test
     @Order(1)
     public void testLastPrice() {
@@ -23,6 +29,6 @@ public abstract class AbstractAMQPTest {
                         .get("/price")
                     .then()
                         .statusCode(200)
-                        .body(equalTo("[10, 20, 30, 40, 50, 60, 70, 80, 90, 100]")));
+                        .body(containsString("10, 20, 30, 40, 50, 60, 70, 80, 90, 100")));
     }
 }
--- a/messaging/artemis-jta/src/main/java/io/quarkus/ts/openshift/messaging/artemisjta/ConsumerService.java
+++ b/messaging/artemis-jta/src/main/java/io/quarkus/ts/openshift/messaging/artemisjta/ConsumerService.java
@@ -1,6 +1,8 @@
 package io.quarkus.ts.openshift.messaging.artemisjta;
 
 import io.quarkus.scheduler.Scheduled;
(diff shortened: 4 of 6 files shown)


messaging/amqp-reactive/src/main/resources/application.properties
```diff
@@ -1,9 +1,6 @@
 quarkus.openshift.expose=true
 quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
 
-# Configures the AMQP broker credentials.
-amqp-username=quarkus
-amqp-password=quarkus
```
Sgitario, line 6: The broker is configured with security disabled, so this configuration is not needed. Moreover, with this configuration the Openshift test was randomly failing.

messaging/amqp-reactive/src/test/java/io/quarkus/ts/openshift/messaging/amqp/AbstractAMQPTest.java
```diff
@@ -23,6 +23,6 @@ public void testLastPrice() {
                         .get("/price")
                     .then()
                         .statusCode(200)
-                        .body(equalTo("[10, 20, 30, 40, 50, 60, 70, 80, 90, 100]")));
+                        .body(containsString("10, 20, 30, 40, 50, 60, 70, 80, 90, 100")));
```
Sgitario, line 26: This solves the issue described in https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/93#issuecomment-720288017.
Tbh, I didn't really like to change the condition from equals to contains. Though it was the easier solution. Another option would be to change the producer to only produce 10 events...

common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
```diff
@@ -53,7 +53,9 @@ public void awaitReadiness(List<HasMetadata> resources) {
                 .forEach(it -> {
                     System.out.println(ansi().a("waiting for ").a(readableKind(it.getKind())).a(" ")
                             .fgYellow().a(it.getMetadata().getName()).reset().a(" to become ready"));
-                    await().atMost(5, TimeUnit.MINUTES).until(() -> {
+                    await().pollInterval(30, TimeUnit.SECONDS)
```
Ladicek, line 56: 30 seconds delay means this will slow down things a _lot_ when the resource is not immediately ready. The default is something like 100 millis or so, which might be too much, but I wouldn't go for more than 1 second.

messaging/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+
+    <name>Quarkus OpenShift TS: Parent: Messaging</name>
+
+    <modules>
+        <module>artemis</module>
+        <module>artemis-jta</module>
+        <module>amqp-reactive</module>
+        <module>qpid</module>
+    </modules>
+</project>
```
Ladicek, line 25: This parent seems totally useless? Big advantage of previous solution is that _all_ modules are listed on one place.

Sgitario: Also, you can still list the modules by running "mvn validate".
From my point of view, having the common modules within a single parent is useful when you want to run a subset only. Moreover, we can delegate common configuration to these subset of modules by having the configuration in its parent.

common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
Sgitario: I agree with you. 30 seconds is too much. Changing to 1 second. Thanks!

messaging/pom.xml
rsvoboda: I prefer having modules defined in on place atm. `messaging/pom.xml` doesn't bring big additional value.

We can introduce it once we have more urgent need for such messaging parent. Similar situation is with https://github.com/quarkus-qe/quarkus-openshift-test-suite/tree/master/sql-db, https://github.com/quarkus-qe/quarkus-openshift-test-suite/tree/master/security etc.

Sgitario:
Found a new issue in Artemis + JTA:

```
Error:  Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 35.67 s <<< FAILURE! - in io.quarkus.ts.openshift.messaging.artemisjta.ArtemisTest
Error:  testPrice  Time elapsed: 7.55 s  <<< ERROR!
org.awaitility.core.ConditionTimeoutException: 
Assertion condition defined as a io.quarkus.ts.openshift.messaging.artemisjta.AbstractArtemisTest 1 expectation failed.
Response body doesn't match expectation.
Expected: "666:666"
  Actual: 666:
 within 5 seconds.
Caused by: java.lang.AssertionError: 
1 expectation failed.
Response body doesn't match expectation.
Expected: "666:666"
  Actual: 666:
```

Investigating...

messaging/pom.xml
Sgitario: Ok, as you both think the same, I've reverted this change.

Sgitario:
> Found a new issue in Artemis + JTA:
> 
> ```
> Error:  Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 35.67 s <<< FAILURE! - in io.quarkus.ts.openshift.messaging.artemisjta.ArtemisTest
> Error:  testPrice  Time elapsed: 7.55 s  <<< ERROR!
> org.awaitility.core.ConditionTimeoutException: 
> Assertion condition defined as a io.quarkus.ts.openshift.messaging.artemisjta.AbstractArtemisTest 1 expectation failed.
> Response body doesn't match expectation.
> Expected: "666:666"
>   Actual: 666:
>  within 5 seconds.
> Caused by: java.lang.AssertionError: 
> 1 expectation failed.
> Response body doesn't match expectation.
> Expected: "666:666"
>   Actual: 666:
> ```
> 
> Investigating...

The issue is in the [ConsumerService.java](https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/0aca538a62df35744dda98faff59cf70db538195/messaging/artemis-jta/src/main/java/io/quarkus/ts/openshift/messaging/artemisjta/ConsumerService.java#L43):

```
try (JMSContext context = connectionFactory.createContext(Session.AUTO_ACKNOWLEDGE)) {
   // read from queue 1
   // read from queue 2
}
```

As the auto acknowledge is configured for both queues and the scheduled jobs can be run concurrently, when the value is read only from queue 1, we ack for both queues and hence the value from queue 2 is gone (in a concurrent run). This can be easily reproduced by adding a sleep in the first execution.

messaging/artemis-jta/src/main/java/io/quarkus/ts/openshift/messaging/artemisjta/ConsumerService.java
```diff
@@ -38,28 +42,14 @@ public void setPrice2(String price) {
...
-            }
-        }
+    @Scheduled(every = "1s", concurrentExecution = ConcurrentExecution.SKIP)
+    public void receiveMessagesInCustomPriceOne() throws JMSException {
+        receiveMessagesInQueue("custom-prices-1", this::setPrice1);
+    }
+
+    @Scheduled(every = "1s", concurrentExecution = ConcurrentExecution.SKIP)
+    public void receiveMessagesInCustomPriceTwo() throws JMSException {
+        receiveMessagesInQueue("custom-prices-2", this::setPrice2);
```
Sgitario, line 52: I made this change to fix the issue in this comment: https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/106#issuecomment-720936901

With concurrent configured to skip, we disallow to trigger the same job if there is a running previous one. Moreover, we avoid to ack messages from two queues at the same time (in receiveMessagesInQueue).

Ladicek (approved):
No messaging expert :-), but this LGTM.

mjurc (approved):
LGTM.

messaging/amqp-reactive/src/test/java/io/quarkus/ts/openshift/messaging/amqp/AbstractAMQPTest.java
rsvoboda: Please add a comment to the AbstractAMQPTest.java saying the same what you have in the above GH comment.

Otherwise good to go!

Sgitario: Added. Let me know if it looks good now. Thanks!

rsvoboda (approved):

rsvoboda:
Merged, thank you!

PR #107, opened 2020-11-02
Print pod logs for failed scenarios

Sgitario:
Changes:
- Allow to configure the command to specify the output (whether a file or the console)
- Update OpenShiftTestExtension to run a list of actions when a scenario fails
- Add two actions to be run after a scenario failure: CopyLogsOnOpenShiftFailureActionImpl and PrintStatusOnOpenShiftFailureActionImpl.

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/Command.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/Command.java
@@ -2,16 +2,22 @@ package io.quarkus.ts.openshift.common;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.List;
+import java.util.function.BiConsumer;
 
 import static org.fusesource.jansi.Ansi.ansi;
 
 public class Command {
     private final String description;
     private final List<String> command;
+
+    private BiConsumer<String, InputStream> outputConsumer = consoleOutput();
 
     public Command(String... command) {
         this(Arrays.asList(command));
@@ -20,6 +26,16 @@ public class Command {
     public Command(List<String> command) {
         this.description = descriptionOfProgram(command.get(0));
         this.command = command;
+    }
+
+    public Command outputToFile(File file) {
+        outputConsumer = fileOutput(file);
+        return this;
+    }
+
+    public Command outputToConsole() {
+        outputConsumer = consoleOutput();
+        return this;
     }
 
     private static String descriptionOfProgram(String program) {
@@ -31,7 +47,6 @@ public class Command {
 
     public void runAndWait() throws IOException, InterruptedException {
         System.out.println(ansi().a("running ").fgYellow().a(String.join(" ", command)).reset());
-
         Process process = new ProcessBuilder()
                 .redirectErrorStream(true)
                 .command(command)
@@ -39,18 +54,37 @@ public class Command {
                 .start();
 
         new Thread(() -> {
-            try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
-                String line;
-                while ((line = reader.readLine()) != null) {
-                    System.out.println(ansi().fgCyan().a(description).reset().a("> ").a(line));
-                }
-            } catch (IOException ignored) {
-            }
+            outputConsumer.accept(description, process.getInputStream());
         }, "stdout consumer for command " + description).start();
 
         int result = process.waitFor();
         if (result != 0) {
             throw new RuntimeException(description + " failed (executed " + command + ", return code " + result + ")");
         }
+    }
+
+    private static final BiConsumer<String, InputStream> fileOutput(File targetFile) {
+        return (description, is) -> {
+            try (OutputStream outStream = new FileOutputStream(targetFile)) {
+                byte[] buffer = new byte[8 * 1024];
+                int bytesRead;
+                while ((bytesRead = is.read(buffer)) != -1) {
+                    outStream.write(buffer, 0, bytesRead);
+                }
+            } catch (IOException ignored) {
+            }
+        };
+    }
+
+    private static final BiConsumer<String, InputStream> consoleOutput() {
+        return (description, is) -> {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    System.out.println(ansi().fgCyan().a(description).reset().a("> ").a(line));
+                }
+            } catch (IOException ignored) {
+            }
+        };
     }
 }
--- a/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
@@ -5,6 +5,7 @@ import io.fabric8.openshift.api.model.ImageStream;
 import io.fabric8.openshift.api.model.Route;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.quarkus.ts.openshift.app.metadata.AppMetadata;
+import io.quarkus.ts.openshift.common.actions.OnOpenShiftFailureAction;
 import io.quarkus.ts.openshift.common.config.Config;
 import io.quarkus.ts.openshift.common.injection.InjectionPoint;
 import io.quarkus.ts.openshift.common.injection.TestResource;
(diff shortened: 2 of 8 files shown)


common/src/main/java/io/quarkus/ts/openshift/common/Command.java
```diff
@@ -1,18 +1,24 @@
...
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.List;
-
-import static org.fusesource.jansi.Ansi.ansi;
+import java.util.function.BiConsumer;
```
Sgitario, line 14: I used the same formatter than Quarkus ... should I revert this and use another? Don't want to create unnecessary changes everywhere.

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
```diff
@@ -22,35 +28,67 @@ public Command(List<String> command) {
...
-                .redirectErrorStream(true)
-                .command(command)
-                .directory(new File(".").getAbsoluteFile())
-                .start();
+            int result = process.waitFor();
+            if (result != 0) {
+                throw new RuntimeException(description + " failed (executed " + command + ", return code " + result + ")");
+            }
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
```
Ladicek, line 66: Any particular reason to wrap exceptions like this?  Whomever calls `runAndWait` should be prepared to deal with interruption.

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
```diff
@@ -450,6 +455,15 @@ public void handleTestExecutionException(ExtensionContext context, Throwable thr
         throw throwable;
     }
 
+    private void runOnFailureAction(ExtensionContext context, OnOpenShiftFailureAction action) {
+        try {
+            postProcessTestInstance(action, context);
```
Ladicek, line 460: This looks weird on the first sight, we should extract the injection logic into a separate, properly named method, so that it's clear what we do here.

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
Ladicek: There's an import sort Maven plugin configured in this project, `mvn compile` should apply I think?

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Sgitario: I agree with you, but we cannot change the postProcessTestInstance name... I can create another proper named method and make the postProcessTestInstance and runOnFailureAction to call to the new one if you prefer so.

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
Sgitario: I will undo this and add the try/catch in the new callers.

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Ladicek: That's what I meant :-)

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
Ladicek: I think the callers can just propagate the exception, too. The generic "on failure action" interface should `throws Exception`, because it can execute arbitrary code.

That way, when error happens, you don't have a terribly long stack trace to go through.

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
Sgitario: I don't mean only about sort import, but general formatter rules. This is the formatter I think we should follow: https://github.com/quarkusio/quarkus/tree/master/independent-projects/ide-config/src/main/resources

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
Sgitario: Not a big fan of having an interface with "throws Exception", I prefer specific exceptions instead. 
On the other hand, these interfaces are to be called when the test scenario fails. So, I guess it's fine to state that the method might throw any exception and this exception will not hide the original root cause of the test failure (it will only print a log message about the exception). 
So, I will change the interface definition and document it properly.

Sgitario: Changed.

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Sgitario: Changed.

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
```diff
@@ -38,16 +39,20 @@
...
 import java.util.Optional;
+import java.util.ServiceLoader;
 import java.util.stream.Stream;
 
 import static org.fusesource.jansi.Ansi.ansi;
 import static org.junit.platform.commons.util.AnnotationUtils.findAnnotatedFields;
 
-// TODO at this point, this class is close to becoming unreadable, and could use some refactoring
+// TODO at this point, this class is close to becoming unreadable, and could use some refactoring. 
+// Raised https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/108 for the refactoring.
```
Sgitario, line 49: As stated in the comment, I've raised the https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/108 issue to do this.

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
rsvoboda: yup, formater inclusion is on our todo, last time @Ladicek had some objections to Quarkus formatter, so we didn't push hard on it. But as more people will be looking into the project, it might be time to push it :)

Ladicek: The Eclipse formatter is crap and more often than not, I have to fight it to leave the code at least a little readable. I hate it with a passion, but feel free to use it on new code :-)

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
Ladicek: I personally think that if the "on failure" action should be able to "do anything", then it naturally needs to be able to "throw anything". See e.g. `java.util.concurrent.Callable` :-)

common/src/main/java/io/quarkus/ts/openshift/common/actions/CopyLogsOnOpenShiftFailureActionImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    public void execute() throws Exception {
+        String namespace = openShiftUtil.getNamespace();
+        for (Pod pod : openShiftUtil.getPods()) {
+            String podName = pod.getMetadata().getName();
+            new Command("oc", "logs", podName).outputToFile(getOutputFile(podName, namespace)).runAndWait();
+        }
+    }
+
+    static String getOutputFolder() {
+        return Config.get().getAsString(INSTANCES_LOGS_OUTPUT_DIRECTORY, DEFAULT_LOG_OUTPUT_DIRECTORY);
```
Ladicek, line 30: 1. You can also inject `Config` with `@TestResource`.
2. The config key must be `ts.*`. The config system will ignore keys that don't begin with `ts.`.

Ladicek:
Just one comment, otherwise LGTM.

common/src/main/java/io/quarkus/ts/openshift/common/actions/CopyLogsOnOpenShiftFailureActionImpl.java
Sgitario: Changed

common/src/main/java/io/quarkus/ts/openshift/common/Command.java
rsvoboda: Whole codebase would get (re)formatted, I would go with https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L459 way not to need to worry about everyone's IDE setup

common/src/main/java/io/quarkus/ts/openshift/common/actions/CopyLogsOnOpenShiftFailureActionImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import io.quarkus.ts.openshift.common.Command;
+import io.quarkus.ts.openshift.common.config.Config;
+import io.quarkus.ts.openshift.common.injection.TestResource;
+import io.quarkus.ts.openshift.common.util.OpenShiftUtil;
+
+import java.io.File;
+
+public class CopyLogsOnOpenShiftFailureActionImpl implements OnOpenShiftFailureAction {
+
+    private static final String INSTANCES_LOGS_OUTPUT_DIRECTORY = "instance-logs";
```
rsvoboda, line 13: @Ladicek wrote `The config key must be ts.*. The config system will ignore keys that don't begin with ts..`
You probably forgot about it.

README should be updated to include info about this property

rsvoboda (changes requested):

common/src/main/java/io/quarkus/ts/openshift/common/actions/CopyLogsOnOpenShiftFailureActionImpl.java
Ladicek: It's in the `Config` javadoc.

rsvoboda: exactly

Sgitario: Good point @rsvoboda . Added.

Ladicek (approved):

rsvoboda (approved):

PR #109, opened 2020-11-03
Run serverless module (surefire part) as part of CI

rsvoboda:
Run serverless module (surefire part) as part of CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           java-version: openjdk${{ matrix.java }}
       - name: Build with Maven
-        run: mvn -fae -V -B clean test
+        run: mvn -fae -V -B clean test -Dinclude.serverless
       - name: Zip Artifacts
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
       - name: Build Quarkus master
         run: git clone https://github.com/quarkusio/quarkus.git && cd quarkus && mvn -B clean install -DskipTests -DskipITs -DskipDocs
       - name: Build with Maven
-        run: mvn -fae -V -B clean test -Dquarkus-core-only
+        run: mvn -fae -V -B clean test -Dquarkus-core-only -Dinclude.serverless
```
Sgitario, line 77: Out of curiosity, why cannot this module be executed as part of the whole suite?

Sgitario (approved):
lgtm, just a question.

.github/workflows/ci.yml
rsvoboda: because at the build phase the extension tries to communicate with OCP server and expects knative / OpenShift Serverless to be installed

And that's also why we end at `test` phase and not `package` phase in GH CI jobs

PR #111, opened 2020-11-06
Fix Artemis test issue

Sgitario:
The issue is that the instance of ArtemisJmsProducer does not have the config injected when the scheduled job is invoked.
These changes try to address this issue by getting the same instance when observing the start event.

It also tries to fix a possible transaction timeout issue in the artemis-jta module.

PR #114, opened 2020-11-18
Update joschi/setup-jdk

rsvoboda:
Update joschi/setup-jdk to avoid GH Action errors like https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/111/checks?check_run_id=1410634310


rsvoboda:
CI works again, merging as this is infrastructure fix and no additional review is really necessary

Issue #29, opened 2020-05-07
Figure out how to share some of YAML files


rsvoboda:
Figure out how to share some of YAML files.
We have duplicates on several places in the repository.

Pointed out in https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/28#issuecomment-625148090

Sgitario:
I can see a couple of straight-forward solutions to manage shared YAML files:

- Using a shared Maven module: 

Good: Easy to use as resources will still be part of the classpath
Bad: Modules will depend on this new module (they depend on a common module already)

- Using a common folder: 

Good: Modules won't depend to any new Maven module
Bad: We'll rely on relative paths to resolve the resources... 

What do you think? @rsvoboda @Ladicek

Ladicek:
I think a shared module is fine. I guess we could actually use the `common` module, which mostly contains the test framework, but can easily contain shared resources as well.

PR #116, opened 2020-11-19
Upgrade RH SSO to 7.4

Sgitario:
Changes:
- Added coverage for RH SSO 7.4 only if the Openshift registry is authenticated
- Keep coverage for RH SSO 7.3 only if the Openshift registry is NOT authenticated
- Move the keycloak resources into the common module.


common/src/main/java/io/quarkus/ts/openshift/common/util/AnnotationUtils.java
```diff
@@ -0,0 +1,32 @@
+package io.quarkus.ts.openshift.common.util;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedElement;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public final class AnnotationUtils {
```
Sgitario, line 10: I had to create this utility because the AnnotatedElement.getAnnotationsByType only retrieves the annotations of the super class if the current class has no annotations.

Ladicek: I know why I don't like inheritable annotations. BTW do you know what's wrong with your `getAllAnnotationsByType`? :-)

Ladicek: Sorry, nothing's wrong, I got confused, I thought you're calling something from JUnit, but you're calling the method itself recursively. That's fine (though using a loop would probably be a bit more efficient :-) ).

(I still think inheritable annotations are harder to read, especially if we depend on order of repeatable annotations, which we do. But technically the code is fine.)

Sgitario:
@Ladicek and @rsvoboda I've moved the keycloak resources into the common module in the latest commit. This change is related to https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/29.

rsvoboda:
Please trigger runs with your branch and added module on our internal Jenkins
Jobs `openshift-test-suite-ci` ++ `openshift-test-suite-ci`

rsvoboda:
`@Inherited` ... I miss a bit more details why this was added, I see it's needed for `AbstractSecurityKeycloakOpenShiftIT` but PR description should provide more context about it.
https://github.com/quarkus-qe/quarkus-openshift-test-suite#deploying-additional-resources should have a note about the inheritance of the annotation.

`@OnlyIfNotConfigured` ... I miss the reason why this is needed.
I want to have RHSSO 7.3 variant executed even if `ts.authenticated-registry` is specified. No need to reduce test coverage.

Sgitario:
> @Inherited



> `@Inherited` ... I miss a bit more details why this was added, I see it's needed for `AbstractSecurityKeycloakOpenShiftIT` but PR description should provide more context about it.
> https://github.com/quarkus-qe/quarkus-openshift-test-suite#deploying-additional-resources should have a note about the inheritance of the annotation.
> 
> `@OnlyIfNotConfigured` ... I miss the reason why this is needed.
> I want to have RHSSO 7.3 variant executed even if `ts.authenticated-registry` is specified. No need to reduce test coverage.

I've removed the `@OnlyIfNotConfigured` annotation. So, we'll be running the SSO 7.3 version always, and also the SSO 7.4 version if we're using a authenticated registry in OpenShift.

I've also updated the README.md and the PR description to give more details about why the `@Inherited` annotation was needed.

Ladicek:
I suggested adding `@OnlyIfNotConfigured` because if we want to have many tests with RH SSO, running them with 2 versions takes double the time, which is already a lot. I don't see it as reducing coverage -- I don't think we intend to "support" more than 1 version of RH SSO; instead, I see it as a way to run the tests, in a somewhat degraded form, without having access to authenticated registry. But your choice, obviously :-)

I still don't like inheritable annotations, especially if we rely on repeatable annotations order, but I've said that enough times I hope :-)

mjurc:
I agree with @Ladicek , we should only run the tests for the supported version

rsvoboda:
`@OnlyIfNotConfigured` ... ok, lets get it back and execute just 7.4 with authenticated registry

`@Inhereted` ... let's do what Ladislav and Michal suggested, no @Inhereted for now, we can live without it. Having "all" config at one place is good to build context.

common/src/main/java/io/quarkus/ts/openshift/common/util/AnnotationUtils.java
Sgitario: I've removed the inheritable annotations and also removed the AnnotationUtils.

Sgitario:
> `@OnlyIfNotConfigured` ... ok, lets get it back and execute just 7.4 with authenticated registry
> 
> `@Inhereted` ... let's do what Ladislav and Michal suggested, no @Inhereted for now, we can live without it. Having "all" config at one place is good to build context.

@rsvoboda I've applied all the changes and also triggered the Jenkins jobs:

- https://quarkus-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/openshift-test-suite-ci/41

The job worked fine and only the RH SSO 7.4 was used as expected.

I also triggered this job: openshift-test-suite-native-ci, but it's failing to locate the binary:

```
14:50:05 oc> ++ ls '/tmp/src//*-runner'
14:50:05 oc> ++ wc -l
14:50:05 oc> ls: cannot access '/tmp/src//*-runner': No such file or directory
14:50:05 oc> + '[' 0 -eq 1 ']'
14:50:05 oc> + echo 'Could not locate a native image. Have you build a native image using mvn -Pnative before?'
14:50:05 oc> Could not locate a native image. Have you build a native image using mvn -Pnative before?
14:50:05 oc> + exit 1
14:50:05 oc> subprocess exited with status 1
14:50:05 oc> subprocess exited with status 1
14:50:06 oc> error: build error: error building at STEP "RUN /usr/libexec/s2i/assemble": exit status 1
```

I don't think this is related to the changes of this PR.

security/keycloak/src/main/resources/application.properties
```diff
@@ -3,8 +3,6 @@
 quarkus.oidc.auth-server-url=http://localhost:8180/auth/realms/test-realm
 quarkus.oidc.client-id=test-application-client
 quarkus.oidc.credentials.secret=test-application-client-secret
-# tolerate 1 minute of clock skew between the Keycloak server and the application
```
rsvoboda, line 6: Why are you removing this? I think @Ladicek added it on purpose because of the time sync issue between servers in cluster

common/src/main/java/io/quarkus/ts/openshift/common/OnlyIfNotConfigured.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The {@code @OnlyIfNotConfigured} annotation can be used to selectively enable or disable certain tests
```
rsvoboda, line 11: Add info about `@OnlyIfNotConfigured` into README.md

rsvoboda (changes requested):

security/keycloak/src/main/resources/application.properties
Sgitario: Because the property `quarkus.oidc.token.expiration-grace` does not exist any longer :/ 
I thought adding the `quarkus.oidc.token.lifespan-grace` property instead, but the tests were consistently working fine.
Any thoughts?

common/src/main/java/io/quarkus/ts/openshift/common/OnlyIfNotConfigured.java
Sgitario: Added

security/keycloak/src/main/resources/application.properties
Ladicek: Yea I think it was renamed, because it wasn't just for expiration. I'm not sure what's the default grace period, but I'd say that tolerating 1 minute difference is still a good idea (in a test!), even though we've fixed the clock skew issues by configuring correct NTP server.

Sgitario: So, do you prefer to get the property back `quarkus.oidc.token.lifespan-grace`?

Sgitario:
@rsvoboda the native job is now working fine: https://quarkus-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/openshift-test-suite-native-ci/17/

security/keycloak/src/main/resources/application.properties
rsvoboda: +1 to have quarkus.oidc.token.lifespan-grace, in our env we have time sync issue solved but the TS runs in multiple envs we do not control and it's safer to have this property set to 1 minute

https://quarkus.io/guides/all-config#quarkus-oidc_quarkus.oidc.token.lifespan-grace
quarkus.oidc.token.lifespan-grace: Life span grace period in seconds. When checking token expiry, current time is allowed to be later than token expiration time by at most the configured number of seconds. When checking token issuance, current time is allowed to be sooner than token issue time by at most the configured number of seconds.

Ladicek: Agree.

Sgitario: Added them back using the new property. Thanks both!

README.md
```diff
@@ -222,7 +222,7 @@ In such case, when you enable retaining resources on test failure, it's best to
 
 ### Enabling/disabling tests
 
-The `@OnlyIfConfigured` annotation can be used to selectively enable/disable execution of tests based on a configuration property.
+The `@OnlyIfConfigured` annotation can be used to selectively enable/disable execution of tests based on a configuration property. On the other hand, the `@OnlyIfNotConfigured` annotation will do the same whether the configuration property does not exist.
```
rsvoboda, line 225:
```suggestion
The `@OnlyIfConfigured` annotation can be used to selectively enable/disable execution of tests based on a configuration property.
Similarly, the `@OnlyIfNotConfigured` annotation can be used to selectively enable/disable the execution of tests based on a configuration property unavailability.
```

rsvoboda (approved):

PR #119, opened 2020-11-25
Fix tests on native build

Sgitario:
In OpenShift tests, at running the command: `oc start-build --from-archive=app.tar.gz`, the build is not extracting the content when the test is running in native mode. 

The workaround is to use: `oc start-build --from-file=target/myfile-runner`. This makes the build worked fine in both jvm and native modes.

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
@@ -154,14 +154,21 @@ final class OpenShiftTestExtension implements BeforeAllCallback, AfterAllCallbac
 
             awaitImageStreams(context, openshiftResources);
 
-            // when generating Kubernetes resources, Quarkus expects that all application files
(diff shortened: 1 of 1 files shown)


rsvoboda:
@Ladicek could you please take a look too?

Ladicek:
OK, I see what's going on -- the `quay.io/quarkus/ubi-quarkus-native-binary-s2i` image expects the native binary to be present directly in `/tmp/src`, but what we do likely causes the binary to be in `/tmp/src/target`. Once again, a consequence of the change from `/deployments` to `/deployments/target`. I really hate that change.

Ladicek (approved):
The change here makes sense, though I don't really like how the native binary detection is hardcoded. But there's probably no better way (we could resort back to `oc start-build --from-dir=target` in native case, but that's arguably a bit worse, because it uploads a lot more than necessary), so... well, reluctant approval :-)

rsvoboda:
I'm running CI job, will add my approval once I see the results there. I'm OK with the code changes.

rsvoboda:
CI has some fails, this change goes in the right direction. Merging. Jose will follow-up with additional investigation and fixes

PR #120, opened 2020-11-25
fix ScalingTest flakiness

pjgg:
https://trello.com/c/3WaoglB4

### DIFF
--- a/scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
+++ b/scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
@@ -3,12 +3,12 @@ package io.quarkus.ts.openshift.scaling;
 import io.quarkus.ts.openshift.common.OpenShiftTest;
 import io.quarkus.ts.openshift.common.injection.TestResource;
 import io.quarkus.ts.openshift.common.util.OpenShiftUtil;
+import io.restassured.response.ValidatableResponse;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
-import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
 import static io.restassured.RestAssured.given;
@@ -16,11 +16,14 @@ import static javax.ws.rs.core.Response.Status.OK;
 import static javax.ws.rs.core.Response.Status.SERVICE_UNAVAILABLE;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 @OpenShiftTest
 public class ScalingOpenShiftIT {
 
     public static final String DC_NAME = "test-scaling";
+    private static final int TIMEOUT_SEC = 60;
+    private static final int DELAY_BETWEEN_REQUEST_MS = 100;
 
     @TestResource
     private OpenShiftUtil openShiftUtil;
@@ -38,28 +41,14 @@ public class ScalingOpenShiftIT {
      */
     @Test
     public void scaleUpTest() {
-        given()
-                .when().get("/scaling")
-                .then()
-                .statusCode(OK.getStatusCode());
+        int REPLICAS = 2;
 
-        openShiftUtil.scale(DC_NAME, 2);
-
-        assertThat(openShiftUtil.countReadyReplicas(DC_NAME)).isEqualTo(2);
-
-        Set<String> uniqueIds = Collections.newSetFromMap(new ConcurrentHashMap<>());
-        await().atMost(1, TimeUnit.MINUTES).untilAsserted(() -> {
-            String uniqueId =
-                    given()
-                            .when().get("/scaling")
-                            .then()
-                            .statusCode(OK.getStatusCode())
-                            .extract().asString();
-            uniqueIds.add(uniqueId);
-
-            assertThat(uniqueIds).hasSize(2);
-        });
+        givenResourcePath("/scaling");
+        whenScaleTo(2);
+        thenCheckReplicasAmount(2);
+        whenMakeRequestTo("/scaling", REPLICAS);
     }
+
 
     /**
      * Workflow:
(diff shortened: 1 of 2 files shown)


scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -37,92 +45,94 @@ public void scaleBack() {
...
      */
     @Test
-    public void scaleToZero() {
-        openShiftUtil.scale(DC_NAME, 0);
+    public void scaleToZero() throws InterruptedException {
+
+        givenResourcePath("/scaling");
+        whenScaleTo(0);
+        thenCheckReplicasAmount(0);
+        IntStream.range(0, 10).forEach(num -> makeHttpScalingRequest("/scaling", SERVICE_UNAVAILABLE.getStatusCode()));
```
Sgitario, line 94: Why do we need to test this endpoint 10 times?

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -37,92 +45,94 @@ public void scaleBack() {
...
-                            .when().get("/scaling")
-                            .then()
-                            .statusCode(OK.getStatusCode())
-                            .extract().asString();
-            uniqueIds.add(uniqueId);
-
-            assertThat(uniqueIds).hasSize(2);
-        });
+    public void scaleUpTest() throws InterruptedException {
+        int expectedUniqueIdsAmount = EXPECTED_SCALE_UP;
```
Sgitario, line 49: The name "expectedUniqueIdsAmount" is confused to me. I think a proper name is just REPLICAS.

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -37,92 +45,94 @@ public void scaleBack() {
...
+                .onOverflow().drop()
+                .map(weakUp -> makeHttpScalingRequest(path, expectedHttpStatus))
+                .map(resp -> resp.extract().asString())
+                .map(resp -> Long.parseLong(resp))
+                .transform()
+                .byDroppingDuplicates()
+                .onFailure().recoverWithItem(-1L);
+
+        uniqueIds.subscribe().with(id -> latch.countDown());
+        completed = latch.await(timeout, TimeUnit.SECONDS);
```
Sgitario, line 128: I think this method is too complicated to follow. If I don't understand it wrongly, what is doing here is to get the distinct timestamps autogenerated by each replica from the GET endpoints. It would be easier to use the Awailability framework:

```java
Set<String> replicas = new HashSet<>();
await().pollInterval(DELAY_BETWEEN_REQUEST_MS, TimeUnit.MILLISECONDS)
          .atMost(TIMEOUT_SEC, TimeUnit.SECONDS)
          .untilAsserted(() -> {
                 replicas.add(makeHttpScalingRequest(path, OK.getStatusCode()).extract().asString());
                 assertEquals(replicas.count(), expectedReplicas);
          });
```

Also, as expectedHttpStatus and timeout arguments are always the same, we could remove them as parameters and use directly the static fields.

Finally, as we're asserting the number of replicas here, we would not need having the completed field and the thenCheckRequestCompleted method.

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -37,92 +45,94 @@ public void scaleBack() {
...
+        whenScaleTo(EXPECTED_SCALE_UP);
+        thenCheckReplicasAmount(EXPECTED_SCALE_UP);
+        whenMakeRequestTo("/scaling", expectedUpUniqueIdsAmount, OK.getStatusCode(), TIMEOUT_SEC);
+        thenCheckRequestCompleted(true);
+
+        int expectedDownUniqueIdsAmount = EXPECTED_SCALE_DOWN;
+        whenScaleTo(EXPECTED_SCALE_DOWN);
+        thenCheckReplicasAmount(EXPECTED_SCALE_DOWN);
+        whenMakeRequestTo("/scaling", expectedDownUniqueIdsAmount, OK.getStatusCode(), TIMEOUT_SEC);
+        thenCheckRequestCompleted(true);
```
Sgitario, line 79: As I stated in my comment in the whenMakeRequestTo, I think this method is not necessary (also the boolean "true" does not provide any context about what is checking exactly).

Sgitario (changes requested):
This PR seems to me like a refactor instead of a fix to solve the ScalingTest flakiness. Can you describe what you found and your solution?

Moreover, I added some comments.

rsvoboda:
@pjgg I agree with comments added by @Sgitario, please adjust the code / respond to the comments.

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -38,91 +43,81 @@ public void scaleBack() {
...
+        assertThat(openShiftUtil.countReadyReplicas(DC_NAME)).isEqualTo(expectedAmount);
+    }
+
+    private void whenMakeRequestTo(String path, int expectedReplicas) {
+        Set<String> replicas = new HashSet<>();
+        await().pollInterval(DELAY_BETWEEN_REQUEST_MS, TimeUnit.MILLISECONDS)
+                .atMost(TIMEOUT_SEC, TimeUnit.SECONDS)
+                .untilAsserted(() -> {
+                    replicas.add(makeHttpScalingRequest(path, OK.getStatusCode()).extract().asString());
+                    assertEquals(replicas.size(), expectedReplicas);
```
Sgitario, line 112: The expected value should be the first argument, otherwise the message in case of assertion error is confusing.

scaling/src/test/java/io/quarkus/ts/openshift/scaling/ScalingOpenShiftIT.java
```diff
@@ -38,91 +43,81 @@ public void scaleBack() {
...
+        int REPLICAS = EXPECTED_SCALE_UP;
+        givenResourcePath("/scaling");
+        whenScaleTo(EXPECTED_SCALE_UP);
+        thenCheckReplicasAmount(EXPECTED_SCALE_UP);
+        whenMakeRequestTo("/scaling", REPLICAS);
+
+        REPLICAS = EXPECTED_SCALE_DOWN;
+        whenScaleTo(EXPECTED_SCALE_DOWN);
+        thenCheckReplicasAmount(EXPECTED_SCALE_DOWN);
+        whenMakeRequestTo("/scaling", REPLICAS);
```
Sgitario, line 73: I personally read much clearer the numbers directly instead of EXPECTED_SCALE_DOWN or EXPECTED_SCALE_UP.

Sgitario (approved):

rsvoboda (approved):

PR #124, opened 2020-11-30
Testing older commit of the quarkus workshops repo

rsvoboda:
Use older commit of the quarkus workshops repo as the it moved to Quarkus 1.9 already


mjurc:
LGTM, but what's with the failing checks?

rsvoboda:
That's because of "old" GHActions script. I don't think it's needed to invest into 1.7 branch too much atm

PR #129, opened 2020-12-04
Await deployment readiness before awaiting route response

Ladicek:


rsvoboda:
CI runs passed for `Quarkus OpenShift TS: Config*` on OCP 4.5 and 4.6

PR #138, opened 2020-12-10
Quarkus 1.10.1.Final - conflicts with openjdk/openjdk-11-rhel7 

pjgg:
Some project doesn't run with Quarkus 1.10.1.Final and openjdk/openjdk-11-rhel7

Ticket: https://trello.com/c/ToonQugd

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
@@ -168,9 +168,11 @@ final class OpenShiftTestExtension implements BeforeAllCallback, AfterAllCallbac
 
             Optional<String> binary = findNativeBinary();
             if (binary.isPresent()) {
+                System.out.println("ECOOOOOOOOOOOOOOO " +  binary.get());
                 new Command("oc", "start-build", getAppMetadata(context).appName, "--from-file=" + binary.get(), "--follow")
                         .runAndWait();
             } else {
+                System.out.println("ECOOOOOOOOOOOOOOO 2 ");
                 // when generating Kubernetes resources, Quarkus expects that all application files
                 // will reside in `/deployments/target`, but if we did just `oc start-build my-app --from-dir=target`,
                 // the files would end up in `/deployments`
--- a/deployment-strategies/quarkus/src/main/resources/application.properties
+++ b/deployment-strategies/quarkus/src/main/resources/application.properties
@@ -1,4 +1,4 @@
 quarkus.application.name=deployment-strategy-quarkus
 
 quarkus.openshift.expose=true
-quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
+#quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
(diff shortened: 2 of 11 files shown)


rsvoboda:
What S2I image is used once `registry.access.redhat.com/openjdk/openjdk-11-rhel7` is removed?

`registry.access.redhat.com/openjdk/openjdk-11-rhel7` is the image for product based use, we can't get rid of it. We need another solution.

There can be yet another change related to the expected location of the deployment, cc @Ladicek 
Some details about changes done in Quarkus 1.9 - https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/98

Ladicek:
This message `ERROR No such file /deployments/sql-db-postgresql-1.0.0-SNAPSHOT-runner.jar` would point to that change, indeed. But I'd guess we would have that fixed by now?

rsvoboda:
it's in, looks like there is a new change in 1.10.something

Sgitario:
So, the problem is that the image "registry.access.redhat.com/openjdk/openjdk-11-rhel7" is no longer compatible with community version 1.10.X, right?

Is there a product version that incorporates the changes in the community version 1.10.x? The latest product version that I see is 1.7.6 and I think it would be good to verify this.

If the above image works with the product version, but not with the community version, I think that the right solution should be to add a profile only for product usage and use the special properties in (like `quarkus.s2i.base-jvm-image`).

rsvoboda:
https://quarkus.io/guides/all-config#quarkus-container-image-openshift_quarkus.s2i.base-jvm-image ... registry.access.redhat.com/ubi8/openjdk-11 gets used when those lines are commented out

On product side we still have requirement to run with registry.access.redhat.com/openjdk/openjdk-11-rhel7 S2I image.

Quarkus developers had to do some changes to OpenShift extension, my bet is on default application location.
They probably forgot about product side ... 

On RHBQ product side the are no product specific patches by default, so upstream Quarkus must work with openjdk-11-rhel7 S2I image as it was working before.

rsvoboda:
```
cd /Users/rsvoboda/git/quarkus/extensions
git grep "/deployments"

container-image/container-image-openshift/deployment/src/main/java/io/quarkus/container/image/openshift/deployment/OpenshiftConfig.java:    @ConfigItem(defaultValue = "/deployments/target/")
container-image/container-image-openshift/deployment/src/main/java/io/quarkus/container/image/openshift/deployment/S2iConfig.java:    @ConfigItem(defaultValue = "/deployments/target/")
container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iConfig.java:    @ConfigItem(defaultValue = "/deployments/")
```

So it looks like `container-image-s2i` is used by DB modules 🤔
The expected path in the container should really be /deployments/target/sql-db-postgresql-1.0.0-SNAPSHOT-runner.jar and not /deployments/sql-db-postgresql-1.0.0-SNAPSHOT-runner.jar as indicated in Trello card

As I wrote above, please read https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/98 to get some context about changes which happened to OpenShift related extensions in 1.9. Maybe something similar / breaking was done in 1.10.1.

rsvoboda:
https://access.redhat.com/documentation/en-us/red_hat_build_of_quarkus/1.7/html-single/deploying_your_quarkus_applications_on_red_hat_openshift_container_platform/index#proc-using-openshfit-extension_quarkus-openshift ... this should work with Quarkus 1.10 / Quarkus master

Try it with ubi image and also rhel7 image

rsvoboda:
I see that in htt2/2 PR `quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7:1.1` gets used

pjgg:
accomplished with #160

PR #143, opened 2020-12-15
Fix serverless tests when using ephemeral namespaces

### LINKED ISSUE #125: Ephemeral namespaces not working for Serverless tests
The Serverless module: https://github.com/quarkus-qe/quarkus-openshift-test-suite/tree/master/deployment-strategies/quarkus-serverless is not working when we run the suite using ephemeral namespace:

```
mvn -B -V clean verify -fae -pl app-metadata/deployment,app-metadata/runtime,common,deployment-strategies/quarkus-serverless -Dts.authenticated-registry -Dinclude.serverless
```

The problem is that the image is pushed by Quarkus OpenShift extension into the OpenShift registry and the namespace is generated when running the tests (not at build time).

Sgitario:
For `deployment-strategies/quarkus-serverless` and `deployment-strategies/quarkus` modules, we delegate to the Quarkus OpenShift extension the deployment into OpenShift. The problem is that this is done at build time. 

The easier approach I could find out is to have a strategy to deploy any OpenShift test (any test case annotated with `OpenShiftTest`). At the moment, there are two existing strategies:
- Embedded strategy (default): what was doing in the OpenShiftTestExtension (pushing the resources directly to OpenShift).
- Manual strategy: basically do nothing and delegate `AppMetadata` annotations.

Nothing new so far, just a bit of refactoring. 

What's is new is the new strategy called `UsingQuarkusPluginDeploymentStrategy` which will use the Quarkus OpenShift extension to deploy either an OpenShift or a Knative application. 

It works fine either for ephemeral or fixed namespaces. Also, for native builds.
This is the mvn command I used to test these changes:

```
mvn -B -V clean verify -pl app-metadata/deployment,app-metadata/runtime,common,deployment-strategies/quarkus,deployment-strategies/quarkus-serverless,external-applications/quarkus-workshop-super-heroes,external-applications/todo-demo-app,http,messaging/artemis,security/keycloak -fae -Dts.use-ephemeral-namespaces=true -Pserverless
```

If you agree with this approach, I will go forward by running the whole suite using the Jenkins job.
cc @Ladicek @rsvoboda


Sgitario:
Updated adding native support for the new strategy.

Sgitario:
Updated to also support running a module with multiple test cases when using ephemeral namespaces.

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
```diff
@@ -0,0 +1,98 @@
...
+ * If the test class is using the {@code UsingQuarkusPluginDeploymentStrategy} strategy, the test framework will use the
+ * Quarkus OpenShift extension to deploy the Quarkus Application by using the property `quarkus.kubernetes.deploy`.
+ *
+ */
+public class UsingQuarkusPluginDeploymentStrategy implements DeploymentStrategy {
+
+    private static final String MVN_COMMAND = "mvn";
+    private static final String PACKAGE_GOAL = "package";
+    private static final String QUARKUS_PLUGIN_DEPLOY = "-Dquarkus.kubernetes.deploy=true";
+    private static final String SKIP_TESTS = "-DskipTests=true -DskipITs=true";
```
Ladicek, line 28: This is passed as 1 argument, even though you clearly want to pass 2. If you really need to pass `-DskipITs`, it needs to be a separate argument.

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final boolean isNativeBuild() {
+        return NATIVE.equals(System.getProperty(QUARKUS_PACKAGE_TYPE));
+    }
+
+    private static final List<String> withNativeBuildArguments() {
+        List<String> args = new ArrayList<>();
+        args.add(withProperty(QUARKUS_PACKAGE_TYPE, NATIVE));
+        args.add(withProperty(QUARKUS_NATIVE_CONTAINER_RUNTIME, System.getProperty(QUARKUS_NATIVE_CONTAINER_RUNTIME, DOCKER)));
+
+        Optional.ofNullable(System.getProperty(QUARKUS_NATIVE_BUILDER_IMAGE))
```
Ladicek, line 80: What's wrong with

```java
String nativeBuilderImage = System.getProperty(QUARKUS_NATIVE_BUILDER_IMAGE);
if (nativeBuilderImage != null) {
    args.add(withProperty(QUARKUS_NATIVE_BUILDER_IMAGE, nativeBuilderImage));
}
```

? :-)

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final String withContainerImageGroup(String namespace) {
+        return withProperty(QUARKUS_CONTAINER_IMAGE_GROUP, namespace);
+    }
+
+    private static final String withKubernetesClientNamespace(String namespace) {
+        return withProperty(QUARKUS_KUBERNETES_CLIENT_NAMESPACE, namespace);
+    }
+
+    private static final String withProperty(String property, String value) {
+        return String.format("-D%s=%s", property, value);
```
Ladicek, line 95: What's wrong with `return "-D" + key + "=" + value;`? :-)

deployment-strategies/quarkus-serverless/pom.xml
```diff
@@ -17,9 +17,6 @@
     <name>Quarkus OpenShift TS: Deployment Strategies Serverless: Quarkus</name>
 
     <properties>
-        <!-- this could also be set in application.properties, but this is much closer to what users actually do -->
-        <!-- We do not use <quarkus.kubernetes.deploy>true</quarkus.kubernetes.deploy> intentionally, just the build -->
-        <quarkus.container-image.build>true</quarkus.container-image.build>
         <quarkus.kubernetes-client.trust-certs>true</quarkus.kubernetes-client.trust-certs>
```
Ladicek, line 20: This should probably also be removed, then? (And possibly added to `UsingQuarkusPluginDeploymentStrategy`.)

Ladicek:
Added a few minor comments, but the main thing is: I really wanted to avoid running Maven from inside the test. This actually overwrites class files that are currently executing, which is a Bad Thing Waiting to Happen.

Obviously I don't see a good way to do this (otherwise it would already be done), just some bad ways that only differ in how bad they are :-)

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
Sgitario: Having the embedded space was correctly interpreted by processbuilder. However, I've separated it into two fields "SKIP_TESTS" and "SKIP_ITS" to avoid misunderstandings ...

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
Sgitario: Changed.

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
Sgitario: What's wrong with `String.format("-D%s=%s", property, value);`? I personally prefer using `String.format`.

deployment-strategies/quarkus-serverless/pom.xml
Sgitario: Well spotted! Moving to UsingQuarkusPluginDeploymentStrategy.

Sgitario:
> Added a few minor comments, but the main thing is: I really wanted to avoid running Maven from inside the test. This actually overwrites class files that are currently executing, which is a Bad Thing Waiting to Happen.
> 
> Obviously I don't see a good way to do this (otherwise it would already be done), just some bad ways that only differ in how bad they are :-)

I disagree that this approach is just another way to do bad things: it covers Quarkus functionality and avoids a lot of compatibility issues among Quarkus versions. 

Of course, I had avoided calling maven inside a test too, but as far I can see it, the root problem is that the test suite is designed to run the tests into OpenShift as if these tests were simple Quarkus tests. This is good for maintainability and readability but implies that the code to cover and the test cases are in the same module. At the moment, there are a lot of internal hacks (where to pick the target files, manual triggers, etc). As part of this PR, we can just use the Quarkus capabilities to deploy into OpenShift. Still hacky though, but an improvement from my point of view. 

Perhaps, a better solution should be to separate the code app and the test cases into separate modules, still the test case should drive the deployment (calling maven), but it would be less troublesome. I didn't do it this way as it would require much more changes and lose readability/maintainability. 

Let me know your thoughts about any other further improvements or more correct solutions.

Ladicek:
> the root problem is that the test suite is designed to run the tests into OpenShift as if these tests were simple Quarkus tests. This is good for maintainability and readability but implies that the code to cover and the test cases are in the same module

Nope, that's not a requirement at all, which you can see in the `external-applications` module. That is because ...

> there are a lot of internal hacks (where to pick the target files, manual triggers, etc)

... manual deployment is not a hack, it's a well-designed feature that nicely fits into the overall design. (The way how we find the target files to deploy is rather ugly, but that has more to do with what Quarkus does and we should be able to simplify that after https://github.com/quarkusio/quarkus/pull/13970.)

(In addition, I'd like to generalize the notion of app metadata, which is currently also very Quarkus specific, but that's for another day and off topic here.)

What doesn't fit at all is Quarkus performing deployment during the Maven `package` phase. That's contrary to the Maven lifecycle design (which is obviously rather rigid...). If only there was something like `mvn quarkus:deploy` (similarly to e.g. Fabric8 Maven plugin or JKube), that would be much better and running that from the test suite wouldn't pose a risk (if it worked incrementally...). I remember there were discussions about that, but I don't think there was ever a conclusion.

OK, getting back to reality. I agree that "something like this" is, in principle, the right thing to do. The present implementation, however, is quite fragile in my opinion, because it overwrites class files of classes that are currently executing.

Here's a silly idea to make ephemeral namespaces always work: we could create a tiny Maven plugin, bound by default to the `initialize` phase, which would check if ephemeral namespaces are enabled (just a system property check) and if so, create one, make it "current" (`oc project ...`) and write its name to a file (say `ts-precreated-namespace`). Later, the test framework would check if such file exists and if so, it wouldn't create an ephemeral namespace again. You could say that's a hack, and you'd be right in a way, but you could also look from a slightly different perspective and see this as the right thing to do. Because ultimately, it's a lifecycle problem, and either deployment or namespace creation has to move.

Sgitario:
> > the root problem is that the test suite is designed to run the tests into OpenShift as if these tests were simple Quarkus tests. This is good for maintainability and readability but implies that the code to cover and the test cases are in the same module
> 
> Nope, that's not a requirement at all, which you can see in the `external-applications` module. That is because ...
> 
> > there are a lot of internal hacks (where to pick the target files, manual triggers, etc)
> 
> ... manual deployment is not a hack, it's a well-designed feature that nicely fits into the overall design. (The way how we find the target files to deploy is rather ugly, but that has more to do with what Quarkus does and we should be able to simplify that after [quarkusio/quarkus#13970](https://github.com/quarkusio/quarkus/pull/13970).)
> 
> (In addition, I'd like to generalize the notion of app metadata, which is currently also very Quarkus specific, but that's for another day and off topic here.)
> 
> What doesn't fit at all is Quarkus performing deployment during the Maven `package` phase. That's contrary to the Maven lifecycle design (which is obviously rather rigid...). If only there was something like `mvn quarkus:deploy` (similarly to e.g. Fabric8 Maven plugin or JKube), that would be much better and running that from the test suite wouldn't pose a risk (if it worked incrementally...). I remember there were discussions about that, but I don't think there was ever a conclusion.
> 
> OK, getting back to reality. I agree that "something like this" is, in principle, the right thing to do. The present implementation, however, is quite fragile in my opinion, because it overwrites class files of classes that are currently executing.
> 
> Here's a silly idea to make ephemeral namespaces always work: we could create a tiny Maven plugin, bound by default to the `initialize` phase, which would check if ephemeral namespaces are enabled (just a system property check) and if so, create one, make it "current" (`oc project ...`) and write its name to a file (say `ts-precreated-namespace`). Later, the test framework would check if such file exists and if so, it wouldn't create an ephemeral namespace again. You could say that's a hack, and you'd be right in a way, but you could also look from a slightly different perspective and see this as the right thing to do. Because ultimately, it's a lifecycle problem, and either deployment or namespace creation has to move.

The plugin idea would not work when having multiple test cases in the same module as each test case should be executed in a different namespace. 

Leaving behind that we have different views about what is a well-designed feature and/or bad solutions, and also because it does not add any value to the discussion. What I tried to address as part of this PR is:

- Fix the modules that used the Quarkus OpenShift extension which was failing when using ephemeral namespaces
- Running multiple test cases in the same module (or when re-running directly via the IDE)

As the Quarkus OpenShift extension is executed at build time, my intention was to use it at test time by running the maven command. 

Benefits:
- We cover the Quarkus OpenShift extension and works fine when using ephemeral namespaces.
- As we reuse all the code from this extension, the version/test framework compatibility is guaranteed -> we don't need to deal with where binaries are, etc.

Issues: 
- It could be troublesome as we're recompiling the same module. 

In order to overcome the above issue, I can think of a few approaches:

a.- Change the build output folder, so we won't recompile the same binaries. 
b.- Having a `mvn quarkus:deploy` command line or a way to deploy programmatically (exposing the OpenShiftProcessor via a new extension). I think this could be even useful for more users for testing purposes. 

a) is easier solution than b), which requires investigation, but b) is more powerful and reusable (for more users).

What do you think?

Ladicek:
Good point re multiple tests in the same module, I didn't think of that.

Also, if Maven skips recompilation, then that's a relief! Sounds like it might be safe then. I'll take another look at the code.

rsvoboda:
rebase pls

Sgitario:
run tests

mjurc:
I'm seeing the following build failures in all `deployment-strategies` modules when the entire test suite is ran:

```
$ mvn -B -V clean verify -Dquarkus-core-only -fae -Dts.authenticated-registry -Dinclude.datagrid -Dinclude.serverless
.
.
.
mvn> [INFO] ------------------------------------------------------------------------
mvn> [INFO] BUILD FAILURE
mvn> [INFO] ------------------------------------------------------------------------
mvn> [INFO] Total time:  10.792 s
mvn> [INFO] Finished at: 2021-01-22T13:32:50Z
mvn> [INFO] ------------------------------------------------------------------------
mvn> [ERROR] Failed to execute goal on project deployment-strategies-quarkus: Could not resolve dependencies for project io.quarkus.ts.openshift:deployment-strategies-quarkus:jar:1.0.0-SNAPSHOT: The following artifacts could not be resolved: io.quarkus.ts.openshift:app-metadata:jar:1.0.0-SNAPSHOT, io.quarkus.ts.openshift:common:jar:1.0.0-SNAPSHOT: Could not find artifact io.quarkus.ts.openshift:app-metadata:jar:1.0.0-SNAPSHOT in libs-snapshot-local (http://<artifactory>) -> [Help 1]
mvn> [ERROR] 
mvn> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
mvn> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
mvn> [ERROR] 
mvn> [ERROR] For more information about the errors and possible solutions, please read the following articles:
mvn> [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
```

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
```diff
@@ -1,30 +1,7 @@
...
-import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
-import org.junit.jupiter.api.extension.ExtensionContext.Store;
-import org.junit.jupiter.api.extension.LifecycleMethodExecutionExceptionHandler;
-import org.junit.jupiter.api.extension.ParameterContext;
-import org.junit.jupiter.api.extension.ParameterResolutionException;
-import org.junit.jupiter.api.extension.ParameterResolver;
-import org.junit.jupiter.api.extension.TestExecutionExceptionHandler;
-import org.junit.jupiter.api.extension.TestInstancePostProcessor;
+import static org.fusesource.jansi.Ansi.ansi;
+import static org.junit.platform.commons.util.AnnotationUtils.findAnnotatedFields;
```
mjurc, line 4: Keep the dependency order :)

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @TestResource
+    private OpenShiftClient openShiftClient;
+
+    @Override
+    public void deploy() throws Exception {
+        String namespace = openShiftClient.getNamespace();
+
+        List<String> args = new ArrayList<>(
+                Arrays.asList(MVN_COMMAND, PACKAGE_GOAL, QUARKUS_PLUGIN_DEPLOY, SKIP_TESTS, SKIP_ITS));
```
mjurc, line 47: We should add `-B -V` here to avoid all the useless download progress logs and on the other hand to see what the Maven is.

mjurc (changes requested):

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Sgitario: I'm using the quarkus formatter and it's being a pain when I modify the source code. Will keep the dependency order.

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
Sgitario: Good idea!

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Ladicek: Hey guys, original author of current import sorting configuration. I feel very strongly about the Quarkus formatter (not in a good way), but I also no longer contribute to this test suite, so feel free to move towards the Quarkus formatter if you wish :-)

Sgitario: I've updated my IDE to not automatically sort the imports, so this issue won't happen to me anymore. 
About the Quarkus formatter, up to the rest of the team

Sgitario: Already fixed.

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
Sgitario: Added.

mjurc:
I'm still seeing the two aforementioned failures mentioned in my previous [comment](https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/143#issuecomment-765427985) after the latest fixes.

Sgitario:
I saw a new error that I didn't experience before:

```
[ERROR] Failed to execute goal on project deployment-strategies-quarkus-serverless: Could not resolve dependencies for project io.quarkus.ts.openshift:deployment-strategies-quarkus-serverless:jar:1.0.0-SNAPSHOT: The following artifacts could not be resolved: io.quarkus.ts.openshift:app-metadata:jar:1.0.0-SNAPSHOT, io.quarkus.ts.openshift:common:jar:1.0.0-SNAPSHOT: Failure to find io.quarkus.ts.openshift:app-metadata:jar:1.0.0-SNAPSHOT in http://10.0.136.110:8040/artifactory/libs-snapshot-local was cached in the local repository, resolution will not be reattempted until the update interval of libs-snapshot-local has elapsed or updates are forced 
```

Investigating ...

Sgitario:
run tests

Sgitario:
run tests

Sgitario:
run tests

Sgitario:
The failures are due to the HTTP module which I think it's unrelated to these changes. 
@pjgg and @mjurc are you aware of these failures? 
Apart from this, I think this PR could be merged.

Sgitario:
Jobs with ephemeral namespaces enabled:
- JVM: https://main-jenkins-csb-quarkusqe.cloud.paas.psi.redhat.com/job/openshift-test-suite-ci/203/
- Native: https://main-jenkins-csb-quarkusqe.cloud.paas.psi.redhat.com/job/openshift-test-suite-native-ci/87/

PR #146, opened 2020-12-15
Verify Http/2 support

pjgg:
### Conclusion (preview):

**Server**: Supports Http/2
**HttpClient**: Doesn't supports Http/2
**RestAssurance**: Doesn't supports Http/2

### How we test it:

**Server**: 

We use an external client that supports Http/2 (Vertx WebClient), and we are checking that the response http version is Http 2.
RestAssurance was discarded, because after several configurations and iterations, we could not make it works. Finally, we found an [open issue](https://github.com/rest-assured/rest-assured/issues/1218) that talks about this. 

**Client**:
We have used Quarkus `quarkus-rest-client` synchronous and asynchronous (mutiny). We made a request, and on the server-side, we recover the HttpRequest protocol version and then return this version as a response header.


http/src/main/resources/application.properties
```diff
@@ -1,5 +1,30 @@
 quarkus.application.name=test-http
-quarkus.http.root-path=/api
+quarkus.kubernetes.readiness-probe.initial-delay=5s
+quarkus.kubernetes.readiness-probe.period=10s
+
 quarkus.kubernetes.deployment-target=openshift
 quarkus.openshift.expose=true
-quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
+#quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
+quarkus.s2i.base-jvm-image=quay.io/jcarvaja/openjdk-11-rhel7
```
rsvoboda, line 8: why ?

http/src/main/resources/application.properties
```diff
@@ -1,5 +1,30 @@
 quarkus.application.name=test-http
-quarkus.http.root-path=/api
+quarkus.kubernetes.readiness-probe.initial-delay=5s
```
rsvoboda, line 2: what's the motivation for this change?

http/src/test/java/io/quarkus/ts/openshift/http/HttpOpenShiftIT.java
```diff
@@ -1,19 +1,14 @@
...
-        when()
-                .get("/hello")
-        .then()
-                .statusCode(200)
-                .body("content", is("Hello, World!"));
+@OnlyIfConfigured("ts.authenticated-registry")
+public class HttpOpenShiftIT extends AbstractHttpTest{
+    @BeforeEach
+    public void setup(){
+        System.setProperty("OpenShiftIT", "true");
```
rsvoboda, line 12: Why is this property set here?

http/src/test/resources/application.properties
```diff
@@ -0,0 +1,30 @@
+
```
rsvoboda, line 1: Why is this file introduced when `http/src/main/resources/application.properties` is also used / available?

rsvoboda (changes requested):

rsvoboda (approved):

rsvoboda:
@Sgitario please review this PR. Feel free to merge it, I have no additional comments to it.

Just for the record I didn't check it against OpenShift instance, I rely on the author that he checked that with the latest changes too.

Sgitario (approved):
lgtm as well

rsvoboda:
Thanks, merging.

Ladicek:
I'm a wee disgruntled with this change. The `http` module was a nice little smoke test that I could use to verify if environment is set up properly, because the test is so basic that nothing can go wrong in it. And now you add HTTP/2 and gRPC to it. IMHO, that should have been a separate module.

rsvoboda:
that's a good point, with this merged we do not have dummy thing to ensure env is not broken (a lot).
@pjgg can you move your changes to dedicated http2 module via new PR?

PR #158, opened 2021-01-18
 Second RHDG scenario - simple failover

mirostary:
1st scenario tests:
- 2 tests - simple get and through JAX-RS endpoint send data into the cache and retrieve it through another JAX-RS endpoint

2nd scenario tests - client failure:
- fill the cache, kill the Quarkus pod, wait for redeployment, and retrieve data
- fill the cache, kill the Quarkus pod, try to send a `put` request, wait for redeployment, and retrieve data

3rd scenario tests - cache failure:
- fill the cache, kill Infinispan cluster, wait for redeployment, and retrieve data
- fill the cache, kill Infinispan cluster, wait for redeployment, retrieve data, and send another `put` request
- fill the cache, kill the Infinispan cluster, try to send a `put` request, wait for redeployment, and retrieve data

Othe changes:
- `InfinispanClientApp` updated with the condition preventing rewriting the cache
-  Updated readme


mirostary:
I can't assign anybody to the PR. @mjurc please take a look at this, thanks.

Ladicek:
You really want the directory structure to mirror the package structure :-)

mirostary:
Thanks, now it should be ok

rsvoboda:
@jsmrcka don't be afraid, for you it fine just to check that you can run those tests. Main review is on @mjurc .

mjurc (changes requested):
This, in my opinion, tests whether the applications can read from cache rather than fail-over and data consistency. I would certainly add some checks for data consistency too:

- add server logic that modifies the data in cache and call endpoint that invokes that, for example an endpoint that functions as a request counter (easiest example would be taking a number from cache and incrementing it by one)
- invoke endpoint on node, make the node fail, check data consistency (like if the increment was processed or no)
- invoke endpoint on failed node, see whether the response of the cluster is what we expect it to be.

infinispan-client/src/main/java/io/quarkus/ts/openshift/infinispan/client/InfinispanFirstGreetingResource.java
```diff
@@ -23,7 +23,7 @@ public String hello() {
         return cache.get("hello");
     }
 
-    @POST
+    @PUT
     @Produces(MediaType.TEXT_PLAIN)
     public String updateHello(String text) {
         cache.put("hello","Hello World, Infinispan is up and changed!");
```
mjurc, line 29: With regards to my review, the change would be `counter++` instead of just `cache.put()`

infinispan-client/src/test/java/io/quarkus/ts/openshift/infinispan/client/InfinispanGreetingResourceOpenShiftIT.java
```diff
@@ -83,6 +92,25 @@ public void testHelloSecondEndpoint() {
...
+                .get("/first-endpoint")
+                .then()
+                .statusCode(200)
+                .body(is("Hello World, Infinispan is up and changed!"));
+
+        when()
+                .get("/second-endpoint")
+                .then()
+                .statusCode(200)
+                .body(is("Hello World, Infinispan is up and changed!"));
```
mjurc, line 111: ...and here, we'd assume the number of requests on counter is what we expect it to be :)

rsvoboda:
@mirostary any news regarding this PR?

mirostary:
@rsvoboda just waiting for a test run to finish and I'll send a new commit ;)

mirostary:
@mjurc Thank you for the review, I hope I did what you meant. I tried to use the counter also without saving it to cache but I wasn't able to access it from tests... I also reduced the number of cluster replicas to 1 for faster launching and to avoid cluster restart problems in the next future tests.

mjurc:
@mirostary Right. What the fail-over test should do is:

- have multiple clients access the data in cache and modify it
- introduce a kind of failure (either in cache or a client)
- verify that the clients and the cache have consistent data

In this case, this means that the test has counter, the apps have counter and the cache has a request counter. This counter increases with each request done to app, and is then pushed into the cache. And even after the failure, the cache should ensure that all clients have consistent data.

mirostary:
@mjurc This PR is ready for review now. Up here, In my first comment, is an updated description of all tests. The multiple client access tests will be in the next PR.

mjurc (approved):
Thanks for PR.

LGTM! :)

rsvoboda:
@mjurc can it be merged? 

@jsmrcka did you look into this PR (see my comment from January 25) ?

mjurc:
I'd say yes, just didn't want to merge it with awaiting review.

jsmrcka (approved):

PR #160, opened 2021-01-18
openjdk/openjdk-11-rhel7 -> ubi8/openjdk-11

pjgg:
- Change deployment folder from /deployments/target to /deployments

- Upgrade Quarkus version to 1.11.0.Final
- move on to ubi-openjdk-11

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
@@ -171,15 +171,7 @@ final class OpenShiftTestExtension implements BeforeAllCallback, AfterAllCallbac
                 new Command("oc", "start-build", getAppMetadata(context).appName, "--from-file=" + binary.get(), "--follow")
                         .runAndWait();
             } else {
-                // when generating Kubernetes resources, Quarkus expects that all application files
-                // will reside in `/deployments/target`, but if we did just `oc start-build my-app --from-dir=target`,
-                // the files would end up in `/deployments`
-                // using `tar` works around that nicely, because the tarball created with this command will have
-                // a single root directory named `target`, which the S2I builder image will unpack to `/deployments`
-                new Command("tar", "czf", "app.tar.gz", "target").runAndWait();
-                new Command("oc", "start-build", getAppMetadata(context).appName, "--from-archive=app.tar.gz", "--follow")
-                        .runAndWait();
-                new Command("rm", "app.tar.gz").runAndWait();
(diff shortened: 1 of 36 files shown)


Sgitario:
run tests

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
```diff
@@ -177,8 +177,7 @@ private void doBeforeAll(ExtensionContext context) throws Exception {
                 // using `tar` works around that nicely, because the tarball created with this command will have
                 // a single root directory named `target`, which the S2I builder image will unpack to `/deployments`
                 new Command("tar", "czf", "app.tar.gz", "target").runAndWait();
-                new Command("oc", "start-build", getAppMetadata(context).appName, "--from-archive=app.tar.gz", "--follow")
-                        .runAndWait();
+                new Command("oc", "start-build", getAppMetadata(context).appName, "--from-dir=target", "--follow").runAndWait();
```
mjurc, line 180: Does this belong to this PR?

mjurc:
Also, perhaps a more informative commit message would be better!

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Ladicek: I think that's required for Quarkus 1.11 support. In which case, we could remove the `tar czf app.tar.gz target` command, that's no longer necessary.

Ladicek: And the comment above, too :-)

sql-db/postgresql/src/main/resources/application.properties
```diff
@@ -1,6 +1,6 @@
 quarkus.openshift.expose=true
-quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
-
+quarkus.s2i.base-jvm-image=registry.access.redhat.com/ubi8/openjdk-11
+
```
Ladicek, line 3: Baaaad whitespace :-)

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
Ladicek: Ah it's already in #163, so it should be removed from here.

mjurc (approved):
The MR now contains updates for 1.11.0, including the changed directory structure for deploment and new base image.

Approving.

It also supersedes #163 now.

PR #164, opened 2021-01-22
qpid flakiness improvements:

pjgg:
* increase awaitability period time, from 1 min to 3 min.
* emit a qpid event every 500 ms (previously was set to 5 sec).
* JMS consumer check if has been received a message before close connection.


messaging/artemis-jta/pom.xml
```diff
@@ -41,7 +41,10 @@
             <groupId>io.quarkus.ts.openshift</groupId>
             <artifactId>app-metadata</artifactId>
         </dependency>
-
+        <dependency>
```
mjurc, line 44: I don't particularly like including commons :) A lot of deps

mjurc (changes requested):

mjurc (approved):
Thanks for MR, LGTM

PR #166, opened 2021-01-22
k8s/openshift secret volumes required a smallrye config env variable …

pjgg:
k8s/openshift secret volumes required a smallrye config env variable location. 
[issue ref](https://github.com/quarkusio/quarkus/issues/14525)

### DIFF
--- a/config-secret/file-system/src/main/resources/application.properties
+++ b/config-secret/file-system/src/main/resources/application.properties
@@ -1,6 +1,9 @@
 quarkus.openshift.expose=true
 quarkus.openshift.secret-volumes.app-config.secret-name=app-config
 quarkus.openshift.mounts.app-config.path=/deployments/config
+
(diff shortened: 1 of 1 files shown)


config-secret/file-system/src/main/resources/application.properties
```diff
@@ -1,6 +1,7 @@
 quarkus.openshift.expose=true
 quarkus.openshift.secret-volumes.app-config.secret-name=app-config
 quarkus.openshift.mounts.app-config.path=/deployments/config
+quarkus.openshift.env.vars.smallrye-config-locations=/deployments/config
```
mjurc, line 4: Please add a TODO to remove the workaround with the link to the issue so we don't keep this in testsuite

mjurc (changes requested):

mjurc (approved):
Thanks, LGTM

PR #176, opened 2021-01-29
Add module for micrometer/prometheus

Sgitario:
Verifies that custom metrics are exposed in the embedded Prometheus instance provided by OpenShift.

There is a PrimeNumberResource that checks whether an integer is prime or not. The application will expose the following custom metrics:
- `prime_number_max_{uniqueId}`: max prime number that was found
- `prime_number_test_{uniqueId}`: with information about the calculation of the prime number (count, max, sum)

Where `{uniqueId}` is an unique identifier that is calculated at startup time to uniquely identify the metrics of the application.

### DIFF
--- a/micrometer/prometheus/src/main/java/io/quarkus/ts/openshift/micrometer/PrimeNumberResource.java
+++ b/micrometer/prometheus/src/main/java/io/quarkus/ts/openshift/micrometer/PrimeNumberResource.java
@@ -1,0 +1,76 @@
+package io.quarkus.ts.openshift.micrometer;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import org.apache.commons.lang3.RandomStringUtils;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import java.util.concurrent.atomic.LongAccumulator;
+import java.util.function.Supplier;
+
+@Path("/")
+public class PrimeNumberResource {
+
+    private final LongAccumulator highestPrime = new LongAccumulator(Long::max, 0);
+    private final MeterRegistry registry;
+    private final String uniqueId = RandomStringUtils.randomAlphabetic(5);
+
+    PrimeNumberResource(MeterRegistry registry) {
+        this.registry = registry;
+
+        // Create a gauge that uses the highestPrimeNumberSoFar method
+        // to obtain the highest observed prime number
+        registry.gauge("prime.number.max." + uniqueId, this,
+                PrimeNumberResource::highestObservedPrimeNumber);
+    }
+
+    @GET
+    @Path("/uniqueId")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String getUniqueId() {
+        return uniqueId;
+    }
+
+    @GET
+    @Path("/check/{number}")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String checkIfPrime(@PathParam("number") long number) {
+        if (number < 1) {
+            return "Only natural numbers can be prime numbers.";
+        }
+        if (number == 1) {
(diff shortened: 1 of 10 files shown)


mjurc (approved):
LGTM!

Sgitario:
run tests

Sgitario:
run tests

Sgitario:
run tests

Sgitario:
run tests

Sgitario:
run tests

micrometer/prometheus/src/main/java/io/quarkus/ts/openshift/micrometer/PrimeNumberResource.java
```diff
@@ -0,0 +1,76 @@
...
+    private final LongAccumulator highestPrime = new LongAccumulator(Long::max, 0);
+    private final MeterRegistry registry;
+    private final String uniqueId = RandomStringUtils.randomAlphabetic(5);
+
+    PrimeNumberResource(MeterRegistry registry) {
+        this.registry = registry;
+
+        // Create a gauge that uses the highestPrimeNumberSoFar method
+        // to obtain the highest observed prime number
+        registry.gauge("prime.number.max." + uniqueId, this,
```
pjgg, line 27: Not sure if its a good idea to create a metric with a Unique Key.  I will suggest use as a key "prime.number.max" and then add tags . So, in this way, we could make queries over "prime.number.max"  and then filter by tag.

micrometer/prometheus/src/main/java/io/quarkus/ts/openshift/micrometer/PrimeNumberResource.java
```diff
@@ -0,0 +1,76 @@
...
+            for (int i = 3; i < Math.floor(Math.sqrt(number)) + 1; i = i + 2) {
+                if (number % i == 0) {
+                    return number + " is not prime, is divisible by " + i + ".";
+                }
+            }
+            highestPrime.accumulate(number);
+            return number + " is prime.";
+        };
+
+        return registry.timer("prime.number.test." + uniqueId).wrap(supplier).get();
```
pjgg, line 65: Add uniqueId as a tag. (sugestion)

Example:

`registry.timer("prime.number" , "DEV", uniqueId)`
`registry.timer("prime.number" , "TEST", uniqueId)`

pjgg (changes requested):

micrometer/prometheus/src/main/java/io/quarkus/ts/openshift/micrometer/PrimeNumberResource.java
```diff
@@ -0,0 +1,76 @@
...
+            return "1 is not prime.";
+        }
+        if (number == 2) {
+            return "2 is prime.";
+        }
+        if (number % 2 == 0) {
+            return number + " is not prime, it is divisible by 2.";
+        }
+
+        Supplier<String> supplier = () -> {
```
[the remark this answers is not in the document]
Sgitario: I wanted to replicate exactly what was done in here: https://quarkus.io/guides/micrometer
This way we can spot issues in the guide over time.

micrometer/prometheus/src/main/java/io/quarkus/ts/openshift/micrometer/PrimeNumberResource.java
Sgitario: Metrics will be deleted after some time from Prometheus. 

Using a unique id will extremely ease the test verification (see how we assert the metrics).

micrometer/prometheus/src/main/java/io/quarkus/ts/openshift/micrometer/PrimeNumberResource.java
Sgitario: Using a unique id extremely ease the test verification (see how we assert the metrics).

pjgg (approved):

PR #179, opened 2021-02-03
Configure env vars from additional deployments. Fixes #127

### LINKED ISSUE #127: Refactor how we populate dependant properties
For example:

```java
// TODO this is pretty ugly, but I'm tired and can't think of a better way at the moment
    @CustomizeApplicationDeployment
    public static void configureKeycloakUrl(OpenShiftClient oc, AppMetadata appMetadata, @WithName("keycloak-plain") URL url) throws IOException {
        keycloakUrl = url + "/auth";
        keycloakRealmUrl = url + "/auth/realms/test-realm";

        List<HasMetadata> objs = oc.load(Files.newInputStream(Paths.get("target/kubernetes/openshift.yml"))).get();
        objs.stream()
                .filter(it -> it instanceof DeploymentConfig)
                .filter(it -> it.getMetadata().getName().equals(appMetadata.appName))
                .map(DeploymentConfig.class::cast)
                .forEach(dc -> {
                    dc.getSpec().getTemplate().getSpec().getContainers().forEach(container -> {
                        container.getEnv().add(
                                new EnvVar("QUARKUS_OIDC_AUTH_SERVER_URL", keycloakRealmUrl, null)
                        );
                    });
                });

        KubernetesList list = new KubernetesList();
        list.setItems(objs);
        Serialization.yamlMapper().writeValue(Files.newOutputStream(Paths.get("target/kubernetes/openshift.yml")), list);
    }
```

Sgitario:
Additional resources with routes sometimes need to be injected into the Quarkus applications. For example, if we have the next `route.yaml` file with content: 

```yaml
apiVersion: v1
kind: List
items:
- apiVersion: route.openshift.io/v1
  kind: Route
  metadata:
    name: keycloak-plain
  spec:
    to:
      name: keycloak-plain
```

Then, We can use the `@InjectRouteUrlIntoApp` annotation to inject the route URL into the Quarkus application as an environment variable:

```java
@OpenShiftTest
@AdditionalResources("classpath:route.yaml")
@InjectRouteUrlIntoApp(route = "keycloak-plain", envVar = "KEYCLOAK_HTTP_URL")
public class HelloOpenShiftIT {
    ...
}
```

After deploying the route, the framework will deploy the module test application with the environment variable `KEYCLOAK_HTTP_URL` and the route URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,42 @@ public class HelloOpenShiftIT {
 These resources are deployed _before_ the test application is deployed, and are also undeployed _after_ the test application is undeployed.
 This annotation is `@Repeatable`, so you can include it more than once.
 
+Also, additional resources with routes sometimes need to be injected into the Quarkus applications. For example, if we have the next `route.yaml` file with content: 
+
+```yaml
+apiVersion: v1
+kind: List
+items:
+- apiVersion: route.openshift.io/v1
+  kind: Route
+  metadata:
+    name: keycloak-plain
+  spec:
+    to:
+      name: keycloak-plain
+```
+
+Then, We can use the `@InjectRouteUrlIntoApp` annotation to inject the route URL into the Quarkus application as an environment variable:
+
+```java
+@OpenShiftTest
+@AdditionalResources("classpath:route.yaml")
+@InjectRouteUrlIntoApp(route = "keycloak-plain", envVar = "KEYCLOAK_HTTP_URL")
+public class HelloOpenShiftIT {
+    ...
+}
+```
+
+After deploying the route, the framework will deploy the module test application with the environment variable `KEYCLOAK_HTTP_URL` and the route URL.
+
+Note that environment variables are available to be used in properties files like:
+
+```properties
+quarkus.my.property=${KEYCLOAK_HTTP_URL:default_value}
+```
+
+Further information about usage of configuration in [here](https://quarkus.io/guides/config-reference#combine-property-env-var).
+
 ### Running tests in ephemeral namespaces
 
 By default, the test framework expects that the user is logged into an OpenShift project, and that project is used for all tests.
--- a/common/src/main/java/io/quarkus/ts/openshift/common/AdditionalResourcesDeployed.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/AdditionalResourcesDeployed.java
@@ -8,6 +8,7 @@ import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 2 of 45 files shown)


rsvoboda:
I would like to get feedback from @mjurc first. Not sure if @Ladicek is willing to look at this too ;)

Ladicek:
I understand the problem needs a proper solution, but this looks like a premature generalization to me (although a really elegant one). I think a simple util class that we could call from a `@CustomizeApplicationDeployment` method would be enough, or if you really want a declarative solution, then I'd add an extra annotation to the test class, something like this: `@InjectRouteUrlIntoApp(route = "keycloak-plain", envVar = "KEYCLOAK_URL")`.

(In fact, I forked the test framework from here into https://github.com/thorntail/openshift-test but didn't really use the opportunity to improve how the SSO booster is tested. For that, I would need a solution to this exact problem. I might take a look :-) )

Sgitario:
> I understand the problem needs a proper solution, but this looks like a premature generalization to me (although a really elegant one). I think a simple util class that we could call from a `@CustomizeApplicationDeployment` method would be enough, or if you really want a declarative solution, then I'd add an extra annotation to the test class, something like this: `@InjectRouteUrlIntoApp(route = "keycloak-plain", envVar = "KEYCLOAK_URL")`.
> 
> (In fact, I forked the test framework from here into https://github.com/thorntail/openshift-test but didn't really use the opportunity to improve how the SSO booster is tested. For that, I would need a solution to this exact problem. I might take a look :-) )

I prefer declarative solutions and liked the idea of having the `@InjectRouteUrlIntoApp` annotation. Also, this way we don't need to dig into the YAML resources to add annotations which is a bit of magic behind scenes. The drawback is that we need another annotation that is tightly coupled with the resources to use.

If you prefer the annotations at test classes approach instead of annotations at resources (or any other), I'm happy to change it. Just let me know. cc @Ladicek @rsvoboda @mjurc

rsvoboda:
I like annotation at test classes a bit more, @mjurc what about you?

rsvoboda:
just thinking, we do deploy the app per `@OpenShiftTest`, right? So we could have different `@InjectRouteUrlIntoApp` mapping for the same envVar in different integration tests (AbcIT.java, CdeIT.java).

Sgitario:
> just thinking, we do deploy the app per `@OpenShiftTest`, right? So we could have different `@InjectRouteUrlIntoApp` mapping for the same envVar in different integration tests (AbcIT.java, CdeIT.java).

Not sure what you mean, but yes. Do you see any issue with this?

Ladicek:
The `@AdditionalResources` are also deployed/undeployed "per" each `@OpenShiftTest`, so that shouldn't pose a problem.

Sgitario:
I've already updated the solution to use the `@InjectRouteUrlIntoApp` annotation. Let me know if it looks better now.

Ladicek (approved):

mjurc (approved):

PR #180, opened 2021-02-03
Testing - Non application endpoints moved to /q/

pjgg:
We've found the following non-application endpoints:
- /metrics
     - /metrics/vendor
     - /metrics/application
    - /metrics/base
- /health
   - /health/live
   - /health/ready
  - /health/well
  - /health/group
- /openapi
- /swagger-ui

Note: paths /q/dev, /q/arc/* doesn't have redirections ( /arc doesn't exist)


jsmrcka:
Is it worth testing different settings for the non-application root path (other than default `/q` path, disabled redirect)? Or would that be unnecessary / beyond the scope? WDYT?

jsmrcka (approved):

pjgg:
> Is it worth testing different settings for the non-application root path (other than default `/q` path, disabled redirect)? Or would that be unnecessary / beyond the scope? WDYT?

As far as I know, all non-application redirections are running under '/q' path.

rsvoboda:
quarkus.http.root-path is fixed in 1.11.1.Final (https://github.com/quarkusio/quarkus/pull/14441)

Please rebase against your branch, master is using 1.11.1.Final now

rsvoboda:
And one more thing, please introduce new module `http-advanced` because the original `http` module was the simplest possible check to ensure cluster / env is setup properly. Now we have more and more thing included into it and I want the simple module back :)

New `http-advanced` should include changes from this PR ++ changes introduced in https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/146

When looking into the history of changes I noticed removal of `quarkus-smallrye-health`. This should get back, we have it in 1.7 branch (https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/1.7/http/pom.xml#L42). It would be Quarkus bug if `quarkus-smallrye-health` is causing troubles.

rsvoboda:
Introduction of `http/http-minimum` will break some jobs we already use / are used by others - 3.11 job and OCP PIT for sure, maybe also IBM Z, IBM P.

@mjurc how do you feel about this move? Just to be clear we can handle 3.11 job and OCP PIT updates, we just don't want to forget about it :)

@pjgg didn't see update of README as part of your PR, but pls wait before we conclude the way forward.
And would use `http-advanced` instead of `http-advance`.

mjurc:
I think the change to `http/http-minimal` and `http/http-advanced` is acceptable.

Please, update `http/http-advance` module to `http/http-advanced` though :) And take care to update all the related infrastructure.

rsvoboda (approved):

PR #183, opened 2021-02-09
Sorted imports by the plugin during the build

rsvoboda:
Sorted imports by the plugin during the build


Sgitario (approved):
Many thanks for this PR! I was really being annoying!

PR #187, opened 2021-02-11
HttpClient redirection coverage

pjgg:


rsvoboda (approved):

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -164,6 +163,29 @@ public void nonAppRedirections() {
         }
     }
 
+    @Test
+    @Disabled("blocked by: QUARKUS-781")
+    public void microprofileHttpClientRedirection(){
+        assertThat(200, equalTo(healthClient.health().getStatus()));
```
Sgitario, line 169: This test will never work when running in OpenShift as `@RestClient` annotation is not injected.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -172,4 +194,16 @@ private static ResponsePredicateResult isHttp2x(HttpResponse<Void> resp) {
         return (resp.version().compareTo(HttpVersion.HTTP_2) == 0) ? ResponsePredicateResult.success() : ResponsePredicateResult.failure("Expected HTTP/2");
     }
 
+    private WebClientOptions defaultVertxHttpClientOptions() throws URISyntaxException {
+        WebClientOptions options = new WebClientOptions().setProtocolVersion(HttpVersion.HTTP_2).setSsl(true).setVerifyHost(false).setUseAlpn(true).setTrustStoreOptions(new JksOptions().setPassword("password").setPath(defaultTruststore()));
+        return options;
```
Sgitario, line 199: Returns the options in the same line.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -172,4 +194,16 @@ private static ResponsePredicateResult isHttp2x(HttpResponse<Void> resp) {
...
+    }
+
+    private String defaultTruststore() throws URISyntaxException {
+        URL res = getClass().getClassLoader().getResource("META-INF/resources/server.truststore");
+        File file = Paths.get(res.toURI()).toFile();
+        String truststore = file.getAbsolutePath();
+        return truststore;
```
Sgitario, line 206: Returns the truststore in the same line.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -97,14 +102,8 @@ public void testGrpc() {
     @Test
     @DisplayName("Http/2 Server test")
     public void http2Server() throws InterruptedException, URISyntaxException {
```
Sgitario, line 104: Rename test case to `vertxHttpClientRedirectionUsingHttp2` and update the `@DisplayName` accordingly for this method and the other too.

Sgitario (changes requested):
Besides my comments line, this test is not compatible when running in OpenShift as we're injecting beans using `@RestClient`.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
pjgg: No, here we are not testing anything related to VertxHttpClient. The goal of this test is to test that server-side support http/2.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
pjgg: ;) Now works!

Sgitario (changes requested):
There are still remaining beans with `@Inject @RestClient`  annotations that do not work when running on native or Openshift mode.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -97,43 +74,49 @@ public void testGrpc() {
...
     @DisplayName("Http/2 Client Sync test")
     @Disabled("blocked by: https://issues.redhat.com/browse/QUARKUS-658")
-    public void http2ClientSync() {
-        Response resp = httpVersionClientService.getClientHttpVersion();
+    public void http2ClientSync() throws Exception {
+        HttpVersionClientService versionHttpClient = HttpVersionClientServiceBuilder.build(getAppEndpoint())
+                .withHostVerified(true)
+                .withPassword(PASSWORD)
+                .withKeyStorePath(KEY_STORE_PATH)
+                .build();
```
Sgitario, line 102: The method names of this pattern are weird: HttpVersionClientServiceBuilder.build()....build();
HttpVersionClientServiceBuilder.withHost(getAppEndpoint()), ... .build() is a more natural choice.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/CommonClientService.java
```diff
@@ -0,0 +1,44 @@
...
+        this.hostVerified = verified;
+        return this;
+    }
+
+    public CommonClientService withKeyStorePath(String keyStorePath) {
+        this.keyStorePath = keyStorePath;
+        return this;
+    }
+
+    public abstract <T> T build() throws Exception;
```
Sgitario, line 33: This is returning a warning because a possible cast exception. It should use generics at class level: CommonClientService<T> and change the implementations accordingly.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/HealthClientServiceBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+
+public class HealthClientServiceBuilder  extends CommonClientService {
+
+    public static HealthClientServiceBuilder build(String uri) {
+        return new HealthClientServiceBuilder(uri);
+    }
+
+    private HealthClientServiceBuilder(String uri) {this.uri = uri;}
```
Sgitario, line 17: The uri field belongs to CommonClientService, so CommonClientService should have a constructor with this field instead.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/HealthClientServiceBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public HealthClientService build() throws CertificateException, NoSuchAlgorithmException, KeyStoreException, IOException {
```
Sgitario, line 19: This code is duplicated in several places. We could move it into CommonClientService.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/CommonClientService.java
```diff
@@ -0,0 +1,44 @@
...
+    protected boolean hostVerified;
+    protected String keyStorePath;
+
+    public CommonClientService withPassword(String password) {
+        this.password = password;
+        return this;
+    }
+
+    public CommonClientService withHostVerified(Boolean verified) {
+        this.hostVerified = verified;
```
Sgitario, line 24: Do we need this field? It's always true...

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/CommonClientService.java
```diff
@@ -0,0 +1,44 @@
...
+        return this;
+    }
+
+    public CommonClientService withKeyStorePath(String keyStorePath) {
+        this.keyStorePath = keyStorePath;
```
Sgitario, line 29: Do we need this field? It's always the same value.

Sgitario (changes requested):
Some cosmetic changes.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/HealthClientServiceBuilder.java
pjgg: it's ok is not exactly the same.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/CommonClientService.java
pjgg: yes, leave it, just in case we have troubles with certificates hostnames.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/CommonClientService.java
pjgg: leave it, just in case we want to test what is happening if you make a request with the wrong certificate.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/clients/HealthClientServiceBuilder.java
Sgitario: I don't see many differences but the last statement: `return builder.build(HealthClientService.class);`

We could move the other code into the parent class and this method to get a Class<T> parameter:

```
public T createService(Class<T> clazz) {
        RestClientBuilder builder = RestClientBuilder.newBuilder().baseUri(URI.create(this.uri));
        builder.hostnameVerifier((s, sslSession) -> this.hostVerified);
        if (!StringUtils.isEmpty(this.keyStorePath)) {
            builder.trustStore(loadKeystore());
        }

        return builder.build(clazz);
```

Sgitario (commented):
I've requested again a few cosmetic changes, but they are not stoppers. 
Up to @pjgg and @rsvoboda to proceed.

Sgitario (changes requested):
After the latest changes, we would only need CommonClientService class (the implementations like HttpVersionClientServiceAsync do not add any value). If this is going to be the final approach, I would also rename the CommonClientService to something like RestClientServiceBuilder or similar.

http/http-advanced/pom.xml
```diff
@@ -59,7 +59,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-micrometer-registry-prometheus</artifactId>
+            <artifactId>quarkus-smallrye-metrics</artifactId>
```
rsvoboda, line 62: `quarkus-micrometer` is the way forward, please use micrometer instead of mp metrics.
You need to add both quarkus-micrometer + quarkus-micrometer-registry-prometheus

rsvoboda (changes requested):

Sgitario (approved):

rsvoboda:
There is fail with Quarkus master:
```
Error:  /home/runner/work/quarkus-openshift-test-suite/quarkus-openshift-test-suite/common/src/main/java/io/quarkus/ts/
 openshift/common/util/ReadinessUtil.java:[16,48] incompatible types: 
 java.lang.Class<capture#1 of ? extends io.fabric8.kubernetes.api.model.HasMetadata>
   cannot be converted to io.fabric8.kubernetes.api.model.HasMetadata

Error:  /home/runner/work/quarkus-openshift-test-suite/quarkus-openshift-test-suite/common/src/main/java/io/quarkus/ts/
  openshift/common/util/ReadinessUtil.java:[23,34] non-static method 
  isReady(io.fabric8.kubernetes.api.model.HasMetadata) cannot be referenced from a static context
```

Sgitario:
> There is fail with Quarkus master:
> 
> ```
> Error:  /home/runner/work/quarkus-openshift-test-suite/quarkus-openshift-test-suite/common/src/main/java/io/quarkus/ts/
>  openshift/common/util/ReadinessUtil.java:[16,48] incompatible types: 
>  java.lang.Class<capture#1 of ? extends io.fabric8.kubernetes.api.model.HasMetadata>
>    cannot be converted to io.fabric8.kubernetes.api.model.HasMetadata
> 
> Error:  /home/runner/work/quarkus-openshift-test-suite/quarkus-openshift-test-suite/common/src/main/java/io/quarkus/ts/
>   openshift/common/util/ReadinessUtil.java:[23,34] non-static method 
>   isReady(io.fabric8.kubernetes.api.model.HasMetadata) cannot be referenced from a static context
> ```

This is caused by this commit: https://github.com/quarkusio/quarkus/commit/db8041c425ca5ef2e9891510e211b83ec34c6509 that was merged yesterday.

Basically, the `isReadinessApplicable` method is protected in fabric8 5.1.0. As a possible workaround, I would suggest copying this method into ReadinessUtil, so we ensure that the test suite works for 1.12 and 1.11.X. 
What do you think?

rsvoboda:
@Ladicek can you please share you view on this? I'm not sure if we can drop `isReadinessApplicable` check from our code.

Ladicek:
This is what I did: https://github.com/thorntail/openshift-test/commit/9474b695cd95f2767401962cf885f58b2b1d3af0

mjurc (approved):
LGTM. Thanks for pull request!

PR #191, opened 2021-02-12
Separate current quarkus version into platform and plugin versions

Sgitario:
Separate current quarkus version into platform (`version.quarkus`) and plugin (`version.plugin.quarkus`) versions. If only platform quarkus version is provided, this version will be used for platform and plugin both. Therefore, we don't need to change anything as it was working previously.

### DIFF
--- a/app-metadata/deployment/pom.xml
+++ b/app-metadata/deployment/pom.xml
@@ -16,18 +16,6 @@
 
     <name>Quarkus OpenShift TS: App Metadata: Deployment</name>
 
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>io.quarkus</groupId>
-                <artifactId>quarkus-bom</artifactId>
-                <version>${version.quarkus}</version>
-                <type>pom</type>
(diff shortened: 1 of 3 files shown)


rsvoboda (commented):
For sure I would pick `Split quarkus version into platform and plugin`

nack for javax.annotation-api workaround

Questions for junit 4
 - how is it possible that we didn't need junit4 before - community rhbq 1.7?
 - what's the failure when junit is not present?

Sgitario:
> For sure I would pick `Split quarkus version into platform and plugin`
> 
> nack for javax.annotation-api workaround
> 
> Questions for junit 4
> 
> * how is it possible that we didn't need junit4 before - community rhbq 1.7?
> * what's the failure when junit is not present?

I've updated the PR to only keep the change about splitting the quarkus versions.

mjurc (approved):
Thanks for PR, looks good to me.

rsvoboda (approved):

rsvoboda:
Waiting for CI

PR #199, opened 2021-02-19
Patch in order to match new fabric8 k8s client signature

pjgg:


Ladicek (approved):
This looks obviously correct, modulo formatting (whitespace between `try{`), but I'd do it like this: https://github.com/thorntail/openshift-test/commit/9474b695cd95f2767401962cf885f58b2b1d3af0#diff-7de6e40917671ea60425d50035108fcf8e20c6e0344bb0364121f595051ea400 That way, the clients don't have to bother with the `.INSTANCE` thing.

Sgitario (approved):

common/src/main/java/io/quarkus/ts/openshift/common/util/ReadinessUtil.java
```diff
@@ -9,21 +9,34 @@
...
+
+    private final Readiness readiness;
+
+    ReadinessUtil(){
+        readiness = new Readiness();
+    }
+
+    public boolean isReadinessApplicable(HasMetadata resource) {
+        try{
+            return readiness.isReady(resource)|| DeploymentConfig.class.isAssignableFrom(resource.getClass());
```
Ladicek, line 28: Ah, I didn't notice this before (and if I did, I wouldn't approve). Calling `readiness.isReady(resource)` just for `isReadinessApplicable` is not a good idea. This method isn't supposed to talk to the Kubernetes API server, it's supposed to be a quick check whether the readiness concept even makes sense for this kind of resource.

Please just abandon the entire idea that you can have a single class that will work with both Fabric8 Kubernetes Client 5.0.x and 5.1.x. They made a breaking change, why pretend they didn't.

Ladicek: Ah one more thing. If I read this correctly, then this method can return `false` for resources where readiness does make sense (and hence the method should return `true`), but are not ready at that particular point in time. This is just bad.

PR #205, opened 2021-02-24
Fix external applications using custom settings.xml

Sgitario:
- Added a custom Maven using RedHat Maven repository and a customisable remote Maven repository
- Configure external applications to use the custom settings and the right Quarkus versions

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
@@ -5,6 +5,7 @@ import io.fabric8.kubernetes.client.Handlers;
 import io.fabric8.kubernetes.client.ResourceHandler;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.quarkus.ts.openshift.app.metadata.AppMetadata;
+import io.quarkus.ts.openshift.common.DefaultTimeout;
 import io.quarkus.ts.openshift.common.OpenShiftTestException;
 
 import java.util.Arrays;
@@ -28,7 +29,7 @@ public final class AwaitUtil {
 
     public void awaitImageStream(String imageStream) {
         System.out.println(ansi().a("waiting for image stream ").fgYellow().a(imageStream).reset().a(" to populate"));
-        await().atMost(5, TimeUnit.MINUTES).until(imageStreamHasTags(oc, imageStream));
+        await().atMost(DefaultTimeout.getMinutes(), TimeUnit.MINUTES).until(imageStreamHasTags(oc, imageStream));
     }
 
     private static Callable<Boolean> imageStreamHasTags(OpenShiftClient oc, String imageStream) {
@@ -45,7 +46,7 @@ public final class AwaitUtil {
 
         System.out.println(ansi().a("waiting for route ").fgYellow().a(metadata.appName).reset()
                 .a(" to start responding at ").fgYellow().a(metadata.knownEndpoint).reset());
-        await().ignoreExceptions().atMost(5, TimeUnit.MINUTES).untilAsserted(() -> {
+        await().ignoreExceptions().atMost(DefaultTimeout.getMinutes(), TimeUnit.MINUTES).untilAsserted(() -> {
             given()
                     // known endpoint is already httpRoot-adjusted
                     .basePath("/")
@@ -64,7 +65,7 @@ public final class AwaitUtil {
                     System.out.println(ansi().a("waiting for ").a(readableKind(it.getKind())).a(" ")
                             .fgYellow().a(it.getMetadata().getName()).reset().a(" to become ready"));
                     await().pollInterval(1, TimeUnit.SECONDS)
-                           .atMost(5, TimeUnit.MINUTES)
+                           .atMost(DefaultTimeout.getMinutes(), TimeUnit.MINUTES)
                            .until(() -> {
                         HasMetadata current = oc.resource(it).fromServer().get();
                         if (current == null) {
--- a/external-applications/quarkus-workshop-super-heroes/src/test/resources/hero.yaml
+++ b/external-applications/quarkus-workshop-super-heroes/src/test/resources/hero.yaml
@@ -23,12 +23,18 @@ items:
       git:
         uri: https://github.com/quarkusio/quarkus-workshops.git
       type: Git
+      configMaps:
+      - configMap:
+          name: settings-mvn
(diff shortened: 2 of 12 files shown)


common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
```diff
@@ -64,7 +64,7 @@ public void awaitReadiness(List<HasMetadata> resources) {
                     System.out.println(ansi().a("waiting for ").a(readableKind(it.getKind())).a(" ")
                             .fgYellow().a(it.getMetadata().getName()).reset().a(" to become ready"));
                     await().pollInterval(1, TimeUnit.SECONDS)
-                           .atMost(5, TimeUnit.MINUTES)
+                           .atMost(10, TimeUnit.MINUTES)
```
Sgitario, line 67: I have to increase this timeout as it tends to spend more time downloading artifacts from artifactory.

common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
```diff
@@ -45,7 +45,7 @@ public void awaitAppRoute() {
 
         System.out.println(ansi().a("waiting for route ").fgYellow().a(metadata.appName).reset()
                 .a(" to start responding at ").fgYellow().a(metadata.knownEndpoint).reset());
-        await().ignoreExceptions().atMost(5, TimeUnit.MINUTES).untilAsserted(() -> {
+        await().ignoreExceptions().atMost(10, TimeUnit.MINUTES).untilAsserted(() -> {
```
Sgitario, line 48: I have to increase this timeout as it tends to spend more time downloading artifacts from artifactory.

common/src/main/resources/deployments/maven/redhat-artifactory-maven-settings.yaml
```diff
@@ -0,0 +1,56 @@
...
+                        <url>https://maven.repository.redhat.com/ga/</url>
+                    </pluginRepository>
+                </pluginRepositories>
+            </profile>
+            <profile>
+                <id>artifactory</id>
+                <repositories>
+                    <repository>
+                        <id>quarkus-maven-repo-zip</id>
+                        <url>http://10.0.136.110:8040/artifactory/quarkus-maven-repo-zip</url>
```
Sgitario, line 34: If the artifactory IP used to change, I can parametrise this if you like. (This would require more changes in the jenkins job too).

external-applications/quarkus-workshop-super-heroes/src/test/resources/hero.yaml
```diff
@@ -23,12 +23,18 @@ items:
...
+          name: settings-mvn
+        destinationDir: "configuration"
     strategy:
       type: Source
       sourceStrategy:
         env:
         - name: ARTIFACT_COPY_ARGS
           value: -p -r lib/ *-runner.jar
+        - name: MAVEN_ARGS
+          value: -s /opt/app-root/src/configuration/settings.xml -Dquarkus-plugin.version=${version.plugin.quarkus} -Dquarkus.platform.version=${version.quarkus} -Dquarkus.platform.group-id=${quarkus.platform.group-id} -Dquarkus.platform.artifact-id=${quarkus.platform.artifact-id} -DskipTests=true
```
Sgitario, line 37: Adding `-s /opt/app-root/src/configuration/settings.xml` is a workaround as the instructions I've found for overwriting the `settings.xml` in OCP didn't work. 

Also I have to skip the tests as they are using docker.

external-applications/quarkus-workshop-super-heroes/src/test/resources/villain.yaml
```diff
@@ -23,12 +23,18 @@ items:
...
+            name: settings-mvn
+          destinationDir: "configuration"
       strategy:
         type: Source
         sourceStrategy:
           env:
             - name: ARTIFACT_COPY_ARGS
               value: -p -r lib/ *-runner.jar
+            - name: MAVEN_ARGS
+              value: -s /opt/app-root/src/configuration/settings.xml -Dquarkus.version=${version.plugin.quarkus} -DskipTests=true
```
Sgitario, line 37: Adding `-s /opt/app-root/src/configuration/settings.xml` is a workaround as the instructions I've found for overwriting the `settings.xml` in OCP didn't work. 

Also I have to skip the tests as they are using docker.

common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
Ladicek: I'd suggest doing something like this: https://github.com/thorntail/openshift-test/commit/0ca7ba4d60ef3742b5a18d6a606ea099db8df589

Sgitario: It makes totally sense. Will add it. Thanks for sharing!

Sgitario: Added.

common/src/main/resources/deployments/maven/redhat-artifactory-maven-settings.yaml
mjurc: I would personally like to see this parametrised.

The test suite is consumed by other parties, and it'd be neat to hide the IP.

mjurc (changes requested):
The Maven remote repository should be parametrised imho.

common/src/main/resources/deployments/maven/redhat-artifactory-maven-settings.yaml
Sgitario: Sure. I've already updated the PR by adding a `ts.s2i.maven.remote.repository` new parameter. Then, we need to update the Jenkins job to add our Artifactory IP.

Sgitario:
@mjurc I addressed your comment. Note that the test works even without providing the new `ts.s2i.maven.remote.repository` property.

rsvoboda:
What's the impact on other users of OpenShift TS - e.g. PIT, IBM P/Z?

rsvoboda:
Usage of internal maven repo brings new requirement for other teams using OpenShiftTS

Conclusions:
 - parameter already documented in readme
 - it works with upstream without need for any parameter

mjurc (approved):
LGTM, thanks for the fix!

PR #215, opened 2021-03-04
Fix external applications using custom settings.xml

Sgitario:
Cherry-pick of https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/205


Sgitario:
About the failures:
- knative is not working fine, I have to dig further about this, but this is not related to this PR
- workshop failed because there were some network issues in the redhat maven repository. Actually, for native it worked fine. If this tend to happen, we should increase the timeout (this can be parametrise now).

@mjurc I think we're ok to merge this PR.

mjurc (approved):

PR #217, opened 2021-03-04
http-advanced: wrong app endpoint path format

pjgg:
We were setting wrong paths that don't match Keycloak disabled patterns 

Example:
`http://localhost:8080/api//hello`

**Note:** there are two slash


rsvoboda:
Why draft? Add some details for reviewers into the description of why is slash expected/required.

mjurc (approved):

Issue #216, opened 2021-03-04
Support propagation of system properties to internal maven command


llowinge:
Using `UsingQuarkusPluginDeploymentStrategy ` and running `mvn clean test -Dquarkus.openshift.expose=true`, the system property is not picked up. It could be probably solved by adding new line `withSystemProperties` after https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/master/common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java#L69.

Sgitario:
Adding this property in the `application.properties` is not an option for you? This is how we use it in `deployment-strategies/quarkus` as an example. 
Our idea is to keep the system properties that need to be passed from arguments to the Quarkus plugin as minimal as possible, the rest of properties should go to `application.properties`.

Let me know if this helps.

llowinge:
Yes, `application.properties` is an option, but i wanted to minimalize number of modifications which needs to be done in some sample quarkus app to be tested on Openshift (eg. modifying `application.properties`). But i understand that you don't want to modify the plugin too much via system properties to avoid some unwanted behaviour. I'm basically struggling only with `quarkus.openshift.expose=true` which is probably mandatory (framework does checks on readiness before it starts tests) unless you want to create it somehow by yourself ? Couldn't be that option be defaulted to true ?

Sgitario:
> Yes, `application.properties` is an option, but i wanted to minimalize number of modifications which needs to be done in some sample quarkus app to be tested on Openshift (eg. modifying `application.properties`). But i understand that you don't want to modify the plugin too much via system properties to avoid some unwanted behaviour. I'm basically struggling only with `quarkus.openshift.expose=true` which is probably mandatory (framework does checks on readiness before it starts tests) unless you want to create it somehow by yourself ? Couldn't be that option be defaulted to true ?

Sure, I will provide a pull request to propagate all the quarkus.* properties to the plugin strategy.

llowinge:
@Sgitario Great, thank you !

rsvoboda:
> Sure, I will provide a pull request to propagate all the quarkus.* properties to the plugin strategy.

I like definition in `application.properties` because it is explicit.

I'm concerned about unwanted side effects of the PR to propagate all the quarkus.* properties to the plugin strategy.

Another approach could be propagation of controlled set of properties (e.g. quarkus.openshift.expose) when they are present.

[...]

PR #219, opened 2021-03-05
Propagate all Quarkus properties when using Quarkus plugin strategy

Sgitario:
When using the Quarkus plugin strategy for deploying into OpenShift, with this PR changes, we'll propagate all the Quarkus properties we enter in the Maven command to the maven command made by the test framework.

As part of the development, I verified the changes for the three scenarios:
- With no parameters
- Using a custom Quarkus argument like `-Dquarkus.openshift.expose=true`
- Running the test in Native

Fixes https://github.com/quarkus-qe/quarkus-openshift-test-suite/issues/216

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
@@ -3,6 +3,7 @@ package io.quarkus.ts.openshift.common.deploy;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.quarkus.ts.openshift.common.Command;
 import io.quarkus.ts.openshift.common.injection.TestResource;
+import org.apache.commons.lang3.StringUtils;
 
 import java.io.IOException;
 import java.nio.file.Files;
@@ -13,6 +14,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
```diff
@@ -121,15 +117,13 @@ private void withMavenRepositoryLocalIfSet(List<String> args) {
...
-    private void withQuarkusBom(List<String> args) {
-        String groupId = System.getProperty(QUARKUS_PLATFORM_GROUP_ID);
-        if (groupId != null) {
-            args.add(withProperty(QUARKUS_PLATFORM_GROUP_ID, groupId));
-        }
-
-        String artifactId = System.getProperty(QUARKUS_PLATFORM_ARTIFACT_ID);
-        if (artifactId != null) {
-            args.add(withProperty(QUARKUS_PLATFORM_ARTIFACT_ID, artifactId));
+    private void withQuarkusProperties(List<String> args) {
```
rsvoboda, line 120: This looks a bit more reader friendly
```java
System.getProperties().entrySet().stream()
      .filter(entry -> StringUtils.startsWith((String)entry.getKey(), QUARKUS_PROPERTY_PREFIX))
      .filter(entry -> StringUtils.isNotEmpty((String)entry.getValue()))
      .forEach(entry -> args.add(withProperty((String)entry.getKey(), (String)entry.getValue())));
```
Otherwise LGTM

Sgitario: Changed to use predicates as you suggested. Let me know if now it looks better.

rsvoboda (approved):

PR #221, opened 2021-03-05
Workshop-super-hero opc additional resources bugfix

pjgg:
Looks that have changed how workshop-super hero packages their application, so we made some build changes accordingly.

Changes:

There is no "-runner.jar" anymore instead we found a rest-hero-01.jar
This, not a über-jar (fatJar), so his dependencies are located on quarkus-app/lib


rsvoboda:
Is it really relevant for 1.11 branch
no "-runner.jar" is since 1.12 imho

pjgg:
that's true! This change doesn't apply for 1.11

PR #224, opened 2021-03-08
Fix http-advanced tests

Sgitario:
The changes made in https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/218 (for Master) and https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/217 (for 1.11) are not working for all the Quarkus versions. To sum up:

- In OpenShift TS using 999-SNAPSHOT: it will work
- In OpenShift TS 1.11 (using 1.11.1.Final): it won't work
- In OpenShift TS 1.11 (setting up the version to 1.11.5.Final): it will work
- In OpenShift TS Master (using 1.11.1.Final): it won't work
- In OpenShift TS Master (using 1.12.0.Final): it won't work
- In OpenShift TS Master (using 1.12.1.Final): it will work

The root cause of the problem is this commit: https://github.com/quarkusio/quarkus/commit/e0f8ec94ef9b6882a60a4f7299c9db9b17893fe2. The changes were back-ported to 1.11.5, but is present in 1.11.1.Final and 1.12.0.Final.

This PR addresses the above inconsistency by adding the slash only if it does not include it and for OpenShift scenario only.

### DIFF
--- a/http/http-advanced/src/main/resources/application.properties
+++ b/http/http-advanced/src/main/resources/application.properties
@@ -53,9 +53,9 @@ quarkus.keycloak.policy-enforcer.paths.health.enforcement-mode=DISABLED
 quarkus.keycloak.policy-enforcer.paths.version.path=/api/httpVersion/*
 quarkus.keycloak.policy-enforcer.paths.version.enforcement-mode=DISABLED
 quarkus.keycloak.policy-enforcer.paths.hello.path=/api/hello/*
-quarkus.keycloak.policy-enforcer.paths.hello.enforcement-mode=DISABLED        
+quarkus.keycloak.policy-enforcer.paths.hello.enforcement-mode=DISABLED
 quarkus.keycloak.policy-enforcer.paths.grpc.path=/api/grpc/*
-quarkus.keycloak.policy-enforcer.paths.grpc.enforcement-mode=DISABLED        
+quarkus.keycloak.policy-enforcer.paths.grpc.enforcement-mode=DISABLED
 
 quarkus.keycloak.policy-enforcer.paths.metrics.path=/api/metrics/*
 quarkus.keycloak.policy-enforcer.paths.metrics.enforcement-mode=DISABLED
--- a/http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
+++ b/http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
@@ -15,6 +15,7 @@ import io.vertx.mutiny.ext.web.client.HttpResponse;
 import io.vertx.mutiny.ext.web.client.WebClient;
 import io.vertx.mutiny.ext.web.client.predicate.ResponsePredicate;
 import io.vertx.mutiny.ext.web.client.predicate.ResponsePredicateResult;
+import org.apache.commons.lang3.StringUtils;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


Sgitario:
run tests

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -177,7 +179,8 @@ public void vertxHttpClientRedirection() throws InterruptedException, URISyntaxE
     }
 
     protected String getAppEndpoint() {
-        return appEndpoint.toString();
+        String endpoint = appEndpoint.toString();
+        return endpoint.endsWith("/") ? endpoint.substring(0, endpoint.length() - 1) : endpoint;
```
pjgg, line 183: I think that is easier to understand the other round way, that is also how the current `master` branch works. 

```
protected String getAppEndpoint() {
        String endpoint = appEndpoint.toString();
        return endpoint.endsWith("/") ? endpoint : endpoint + "/";
    }
```

And all the other changes are gone, with this approach.

pjgg (changes requested):
Could you add a commit message that explains the changes? something like the PR.

mjurc (changes requested):
The code looks good to me, thanks for the fixes!

Like Pablo mentioned, please add better commit message.

Sgitario:
Comments have been addressed. Can you take another look?

http/http-advanced/src/main/resources/application.properties
```diff
@@ -53,9 +53,9 @@ quarkus.keycloak.policy-enforcer.paths.health.enforcement-mode=DISABLED
 quarkus.keycloak.policy-enforcer.paths.version.path=/api/httpVersion/*
 quarkus.keycloak.policy-enforcer.paths.version.enforcement-mode=DISABLED
 quarkus.keycloak.policy-enforcer.paths.hello.path=/api/hello/*
-quarkus.keycloak.policy-enforcer.paths.hello.enforcement-mode=DISABLED        
+quarkus.keycloak.policy-enforcer.paths.hello.enforcement-mode=DISABLED
```
pjgg, line 56: Could you remove the formatting changes on the lines that have not changed?

pjgg (changes requested):

http/http-advanced/src/main/resources/application.properties
Sgitario: This was not an automatic format change, there were space characters here which is wrong.

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -74,7 +75,8 @@ public void testGrpc() {
     @DisplayName("Http/2 Server test")
     public void http2Server() throws InterruptedException, URISyntaxException {
         CountDownLatch done = new CountDownLatch(1);
-        Uni<JsonObject> content = WebClient.create(Vertx.vertx(), defaultVertxHttpClientOptions()).getAbs(getAppEndpoint() + "hello")
+        Uni<JsonObject> content = WebClient.create(Vertx.vertx(), defaultVertxHttpClientOptions())
```
pjgg: remove code format...

http/http-advanced/src/test/java/io/quarkus/ts/openshift/http/AbstractHttpTest.java
```diff
@@ -162,7 +164,8 @@ public void microprofileHttpClientRedirection() throws Exception {
     @Test
     public void vertxHttpClientRedirection() throws InterruptedException, URISyntaxException {
         CountDownLatch done = new CountDownLatch(1);
-        Uni<Integer> statusCode = WebClient.create(Vertx.vertx(), defaultVertxHttpClientOptions()).getAbs(getAppEndpoint() + "health")
+        Uni<Integer> statusCode = WebClient.create(Vertx.vertx(), defaultVertxHttpClientOptions())
```
pjgg: remove code format...

pjgg (approved):

PR #230, opened 2021-03-09
Add swagger-ui to non application redirection test

pjgg:

PR #232, opened 2021-03-10
Feat/backport non app endpoint swagger UI

pjgg:
backport: #230


Sgitario (approved):

rsvoboda (changes requested):
I don't like 3 commits in this PR, it should be just one for the cherry-picked commit.

rsvoboda (approved):

PR #235, opened 2021-03-12
[master] Use absolute path for Maven settings in S2I builds

Sgitario:
The path `/opt/app-root/src` depends on the image that OCP uses to build. In OCP 4+ and 3.11, the working dir differs:
- For OCP 3, the path is `/tmp/src`
- For OCP 4+, the path is `/opt/app-root/src`

In order to support both platforms, this PR aims to use absolute path instead.
I've verified that this is now working with OCP 3 and OCP 4.

This PR needs to be backported to 1.11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,7 +428,7 @@ spec:
     configMaps:
     - configMap:
         name: settings-mvn
-      destinationDir: "configuration"
+      destinationDir: "/configuration"
   strategy:
     type: Source
     sourceStrategy:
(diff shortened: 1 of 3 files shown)


rsvoboda:
Interesting, where can I find the info about the path defaults in the images?

Sgitario:
> Interesting, where can I find the info about the path defaults in the images?

I could not find any place in the documentation to state what the working directory is. Though I found some links where this is stated:
- For OCP 3: https://access.redhat.com/documentation/en-us/red_hat_jboss_middleware_for_openshift/3/html-single/red_hat_java_s2i_for_openshift/index
- For OCP 4: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.3/html-single/images/index

And, in this another link: https://www.openshift.com/blog/create-s2i-builder-image it clearly says that the working dir depends on the image:

```
The destination in the above command (./) is using working directory set in the openshift/base-centos7 image, which is set to be /opt/app-root/src.
```

mjurc (approved):

PR #238, opened 2021-03-16
Non-application endpoint could skip keycloak-authz, by disabling `/q/*`

pjgg:
keycloak-authz: Non-application endpoint could be skipped by disabling `/q/*`
http-advanced: we still skip all non-application endpoints because we want to force a redirection.


mjurc (approved):
LGTM, thanks for updates!

rsvoboda:
@pjgg will similar changes be applied to 1.11 branch?

PR #245, opened 2021-03-18
 Explicitly making app-metadata/runtime use versioned quarkus-bootstrap-maven-plugin.

mjurc:
Fixing the issue when sometimes a wrong version of `quarkus-bootstrap-maven-plugin` is used with unpopulated repositories.

Upstream https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/244


rsvoboda:
Run with master still fails. We have fix for released version, but still pending problem for Q master.
I'm running locally to double check and will merge once the build finishes.

rsvoboda (approved):

PR #251, opened 2021-03-26
Remove executions section in quarkus-bootstrap-maven-plugin

Sgitario:
The executions section is making the build of the app module to fail when using quarkus 999-SNAPSHOT.
Looking at Quarkus upstream, they are not using this executions section.


Sgitario:
The failure is caused by an unrelated issue. Fixed as part of https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/252.

Sgitario:
This solution is not working, the modules failed with:
```
java.lang.RuntimeException: target/app-metadata.properties not found, did you add the app-metadata extension?
Caused by: java.nio.file.NoSuchFileException: target/app-metadata.properties
```

PR #253, opened 2021-03-26
Readiness probe adjustment.

pjgg:
Currently, we are not setting `quarkus.openshift.readiness-probe.period` so the default value (30 sec) is getting in place.
This default value add some extra time that is not needed, especially on test that is increasing and decreasing the number of pods.


Sgitario:
run tests

Sgitario (approved):

rsvoboda:
Is there a way to make these settings global? Having these new entries in every application.properties doesn't look so nice.

PR #252, opened 2021-03-26
Fix grpc port collision on test

Sgitario:
We had some port conflicts in OpenShift TS related to gRPC.
This change is based on https://github.com/quarkus-qe/beefy-scenarios/pull/153 and it should not affect OpenShift tests.

PR #254, opened 2021-03-26
App Metadata needs runtime dependencies of vertx-http

Sgitario:
At the moment, the bootstrap maven plugin is failing when using 999-SNAPSHOT version.

### DIFF
--- a/app-metadata/runtime/pom.xml
+++ b/app-metadata/runtime/pom.xml
@@ -19,7 +19,7 @@
     <dependencies>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-core</artifactId>
+            <artifactId>quarkus-vertx-http</artifactId>
         </dependency>
     </dependencies>
 


Sgitario:
run tests

Sgitario:
The failure is caused by an unrelated issue. Fixed as part of #252.

Sgitario:
The failures in Native are due to timeouts when building in OpenShift. I think we should increase the timeout for Native compilation.
However, this is unrelated to this PR and I think we're good to merge.
cc @rsvoboda

rsvoboda:
@Sgitario were you able to identify PR or issue which brought this change?

Sgitario:
> @Sgitario were you able to identify PR or issue which brought this change?

I tried to identify what changed was the root cause of this new behaviour, but no luck so far.

Sgitario:
@rsvoboda I've asked in zulip just in case anybody can help to clarify this: https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Custom.20Extensions.20now.20needs.20Vertx.20Http.20as.20a.20dependency

Sgitario:
@rsvoboda Georgios has already helped to clarify why started failing. It was an enhancement of the validation: https://github.com/quarkusio/quarkus/pull/14609
I think we're good to merge this change then.

Ladicek (approved):

Ladicek:
Agree, this is the correct thing to do.

rsvoboda (approved):

PR #255, opened 2021-03-26
Deploy in parallel several resources.

pjgg:
Goal: reduce test suite execution time.
Desc: Deploy in parallel additional resources

Nowadays we are deploying in Openshift-ts following a blocking sequential steps:

```
CreateEphimeralNs -> delpoyAdditionalResources(you could have more than one and all of them are blocking) ->  runPublicStaticVoidMethods -> deploy application -> setupRestAssured -> readinesProbe
```

Porposal:

Block One: setup ecosystem
```
CreateEphimeralNs ->
                       - delpoyAdditionalResources
                                           - delpoyAdditionalResources 1
                                           - delpoyAdditionalResources 2
                                           - delpoyAdditionalResources 3
                      - runPublicStaticVoidMethods
```

Block Two(Depends on block one): setup application

```
deployApplication -> setup Rest assured -> readinesProbe
```

Also, you will have a little summary of deployed additional resources
Example:

```
--- Summary: deployAdditionalResources ---
classpath:deployments/kafka/amq-streams.yaml ✓
classpath:deployments/kafka/apicurio.yaml ✓
--- end ---
```

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
@@ -14,6 +14,9 @@ import io.quarkus.ts.openshift.common.injection.WithName;
 import io.quarkus.ts.openshift.common.util.AwaitUtil;
 import io.quarkus.ts.openshift.common.util.OpenShiftUtil;
 import io.restassured.RestAssured;
+import org.apache.commons.lang3.tuple.MutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.fusesource.jansi.Ansi;
 import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
@@ -39,10 +42,15 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.LogManager;
 import java.util.logging.Logger;
@@ -139,18 +147,13 @@ final class OpenShiftTestExtension implements BeforeAllCallback, AfterAllCallbac
 
     private void doBeforeAll(ExtensionContext context) throws Exception {
         System.out.println("---------- OpenShiftTest set up ----------");
-
         createEphemeralNamespaceIfNecessary(context);
-
-        deployAdditionalResources(context);
-
+        CompletableFuture.allOf(deployAdditionalResources(context)).join();
         runPublicStaticVoidMethods(CustomizeApplicationDeployment.class, context);
 
         Map<String, String> envVars = resolveEnvVars(context);
         getDeploymentStrategy(context).deploy(envVars);
-
         setUpRestAssured(context);
-
         getAwaitUtil(context).awaitAppRoute();
     }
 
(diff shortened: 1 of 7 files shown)


Ladicek:
At the very least, you should update this: https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/main/common/src/main/java/io/quarkus/ts/openshift/common/AdditionalResources.java#L11-L12 Because with this proposal, deployment order is no longer guaranteed. Which I personally consider a bad idea, because enforcing serial order is the easiest way of handling dependencies, but if you have measurements showing that this change helps significantly, there are probably other ways.

pjgg:
> At the very least, you should update this: https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/main/common/src/main/java/io/quarkus/ts/openshift/common/AdditionalResources.java#L11-L12 Because with this proposal, deployment order is no longer guaranteed. Which I personally consider a bad idea, because enforcing serial order is the easiest way of handling dependencies, but if you have measurements showing that this change helps significantly, there are probably other ways.

This PR allows you to parallelize additional resources but doesn't stop you to guarantee pod ordering and uniqueness through StatefulSets. 

Anyway, we have added an additional property called `parallel` which default value is going to be false. So by default, the behavior is going to be as it's now (where the ordering relies on an imperative blocking programming, where all the deployments are done in sequence, thanks to is running in a single blocking thread). But, if you want to deploy all your resources at a once (even if they require ordering) and also be sure that if scales, the process is done as you expected, then you can set the property `parallel` to true

Ladicek:
I only meant `oc apply` ordering, not pod ordering. I don't think the Javadoc of `@AdditionalResources` should mention StatefulSets or anything like that -- that's a completely different kind of ordering that is achieved through completely different means.

Note that we provide significant guarantees here (awaiting readiness of the 1st set of deployed resources before starting to deploy a 2nd set of resources). I think that shouldn't be given up lightly. Now I understand and agree that proper "Kubernetes-native" applications shouldn't care about deployment order and should eventually converge to a working state -- but this is a test suite, and enforcing serial deployment order lets you avoid all kinds of hard-to-debug troubles.

Opting into parallel deployment of additional resources makes sense to me -- but please don't add an attribute to `@AdditionalResources`, because then you have to figure out what happens when you have

```java
@AdditionalResources(value = "foo.yaml", parallel = false)
@AdditionalResources(value = "bar.yaml", parallel = true)
```

How about a simple marker `@ParallelAdditionalResources`?

Ladicek:
Also, another option how to deploy resources "in parallel" is to put them into a single YAML file :-) (Which leads to a question why don't we deploy all YAML files at once and await readiness only at the very end. Well, see comment above -- this is a test suite. We don't want to debug other products, Quarkus is complex enough.)

Also, please add something to the README. Thanks!

Sgitario:
> Also, another option how to deploy resources "in parallel" is to put them into a single YAML file :-) (Which leads to a question why don't we deploy all YAML files at once and await readiness only at the very end. Well, see comment above -- this is a test suite. We don't want to debug other products, Quarkus is complex enough.)
> 
> Also, please add something to the README. Thanks!

+100 I agree with @Ladicek . Taking into account the behaviour, I think it's much easier to follow how is behaving right now (sequential order always) instead of doing some parallel magic in behind. About the code in the logic, it's also much harder to follow now. As Ladislav suggested, applying the YAML files at once and then await for readiness for each at the very end would do the same and much simpler.

pjgg:
> l" is to put them into a s

If we move on everything to a single YAML, that doesn't mean that everything is going to be deployed in parallel. Indeed in an StatefulSets everything is going to go in order. The point that we are trying to achieve with this PR is to decouple components. So if you need a Kafka + Zookeeper, a postgresql and a redis, you could define an  `StatefulSets` with the first two-component and an two extra YAML with the other two components, and deploy these three YAML at the same time.

pjgg:
In cases that for example you don't have the docker images that you required in your environment, you could save a significant amount of time. And If you don't want to use it, because you don't want to guarantee uniqueness, ordering and also you are not worried about scaling and unemployment (lose data, consistency etc)...you can leave as it is now!. You don't have to touch any configuration.

Ladicek:
> If we move on everything to a single YAML, that doesn't mean that everything is going to be deployed in parallel.

Please note I'm only talking about the TS behavior. So, whey I say "deploy", I mean "deploy Kubernetes resources", or `oc apply`. What happens in parallel or serially in Kubernetes, that's a wholly different matter.

pjgg:
Ok, let's recap!

 - Instead of creating a new property over the existing annotation `@AdditionalResources` we are going to create a new annotation called `@ParallelAdditionalResourcesEnabled`

- Remove comments related to `StatefulSets` because the current implementation doesn't stop you to run a YAML that has an `StatefulSets` that guarantee deploy/undeploy/scale in order.

Ladicek:
Just noticed you also run the `@CustomizeApplicationDeployment` methods in parallel with `@AdditionalResources`. I'm not sure what I think about that, but if it's intentional, please make a note in the javadoc. And please add something to the README.

pjgg:
> Just noticed you also run the `@CustomizeApplicationDeployment` methods in parallel with `@AdditionalResources`. I'm not sure what I think about that, but if it's intentional, please make a note in the javadoc. And please add something to the README.

Updated!. `@CustomizeApplicationDeployment` was removed from the parallel deployment and we have added a note on the readme to explain what `@ParallelAdditionalResourcesEnabled` try to accomplish.

common/src/main/java/io/quarkus/ts/openshift/common/OpenShiftTestExtension.java
```diff
@@ -274,26 +319,32 @@ private void dropEphemeralNamespaceIfNecessary(ExtensionContext context) throws
...
     }
 
-    private void runPublicStaticVoidMethods(Class<? extends Annotation> annotation, ExtensionContext context) throws Exception {
-        for (Method method : context.getRequiredTestClass().getMethods()) {
-            if (method.getAnnotation(annotation) != null) {
-                if (!isPublicStaticVoid(method)) {
-                    throw new OpenShiftTestException("@" + annotation.getSimpleName()
-                            + " method " + method.getDeclaringClass().getSimpleName() + "." + method.getName()
-                            + " must be public static void");
+    private void runPublicStaticVoidMethods(Class<? extends Annotation> annotation, ExtensionContext context) {
```
Ladicek, line 322: The change here will just lead to harder-to-understand exception messages. Since you no longer run this in parallel with additional resources, can you revert this method back to how it looked?

Ladicek (approved):
Okay then.

PR #268, opened 2021-04-23
Adding onBackpressureDrop for amqp-reactive module

Sgitario:
This test was randomly failing due to the events were not fired.
According to this issue: https://github.com/smallrye/smallrye-reactive-messaging/issues/132#issuecomment-502560543, this can happen when sending too many messages and the way to fix it is either dropping messages on back pressure or using `waitForWriteCompletion(true)`.


Sgitario:
The problem of dropping events is that the expected values order is not guaranteed (1, 2, 3, 4,... ), but 2 can be dropped. Therefore, I've updated the assertion to verify the expected prices in any order.

pjgg (approved):

PR #283, opened 2021-05-07
Resolve retrofit pom version conflict

pjgg:
This commit #267 has introduced a conflict on retrofit library.  Looks that quarkus-universe-bom overwrites retrofit:2.5.0 lib, and apicurio registry (used by kafka-avro-reactive-messaging module) needs retrofit 2.9.0. 

This conflict is fixed on Quarkus 2.X, so this patch is only valid for the previous versions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,7 @@
         <version.quarkus>1.13.3.Final</version.quarkus>
         <version.plugin.quarkus>${version.quarkus}</version.plugin.quarkus>
         <version.testcontainers>1.15.3</version.testcontainers>
+        <version.com.squareup.retrofit2>2.9.0</version.com.squareup.retrofit2>
         <version.formatter-maven-plugin>2.15.0</version.formatter-maven-plugin>
         <version.impsort-maven-plugin>1.6.0</version.impsort-maven-plugin>
         <version.xml-format-maven-plugin>3.1.2</version.xml-format-maven-plugin>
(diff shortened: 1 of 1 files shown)


rsvoboda (changes requested):
Put it into dependency management, no explicit versions in modules pls.
https://github.com/quarkus-qe/quarkus-openshift-test-suite/blob/main/pom.xml#L86

rsvoboda:
@pjgg reminder about description of the PR ... https://gitlab.cee.redhat.com/quarkus-qe/signpost/-/blob/main/pull-requests.md#main-rule-set-for-pull-requests

Sgitario (changes requested):
Same comment as @rsvoboda .

pom.xml
```diff
@@ -161,6 +162,11 @@
                 <artifactId>jansi</artifactId>
                 <version>${version.fusesource-jansi}</version>
             </dependency>
+            <dependency>
```
Sgitario, line 165: Please, add the comment here: 
```
<!-- TODO: com.squareup.retrofit2:retrofit:2.9.0 should be removed on Quarkus 2.0.0.Alphas-->
```

messaging/kafka-avro-reactive-messaging/pom.xml
```diff
@@ -10,6 +10,11 @@
     <artifactId>kafka-avro-reactive-messaging</artifactId>
     <name>Quarkus OpenShift TS: Messaging: Reactive Kafka with Avro</name>
     <dependencies>
+        <!-- TODO: com.squareup.retrofit2:retrofit:2.9.0 should be removed on Quarkus 2.0.0.Alphas-->
+        <dependency>
+            <groupId>com.squareup.retrofit2</groupId>
+            <artifactId>retrofit</artifactId>
+        </dependency>
```
Sgitario, line 17: We don't need to explicitly add this dependency here. Adding it into the dependency management is enough.

Sgitario (changes requested):

common/src/main/java/io/quarkus/ts/openshift/common/deploy/UsingQuarkusPluginDeploymentStrategy.java
```diff
@@ -56,6 +55,58 @@
     @TestResource
     private OpenShiftClient openShiftClient;
 
+    private static final void withQuarkusVersions(List<String> args) {
```
Sgitario, line 58: These format changes are unexpected, can you check why these were introduced?

Sgitario (changes requested):

Sgitario (approved):

PR #297, opened 2021-05-26
Enable test for Keycloak 7.4 and Quarkus Authorization extension

Sgitario:
This is supposed to work in 999-SNAPSHOT and 1.11.7.DR1 (product), but not in 1.11.7.Final (community)


rsvoboda (approved):
I'm ok with doing this change. @pjgg fyi we are re-enabling tests for rhsso. @Sgitario please prepare backport for 1.11 branch

PR #321, opened 2021-08-02
Fix configuration mistake leading to failures in MultiplePersistenceUnitTest

yrodiere:
The test was failing simply because we were telling Quarkus to use a MariaDB dialect, while using an H2 database.
I suppose it used to work with Quarkus 1.x because for some reason the dialect was being ignored, or because the H2 dialect was close enough to the MariaDB dialect. In any case, that was fixed in Quarkus 2.0, so now the invalid dialect configuration is a problem and must be fixed.

See also https://github.com/quarkusio/quarkus/issues/15836.


rsvoboda:
@Sgitario fyi

rsvoboda:
Approved Yoann to run the CI jobs, Yoann is Qaurkus contributor.

Sgitario:
> @Sgitario fyi

This issue was already fixed in the new test suite, so nothing to be migrated.

rsvoboda:
We should still fix quarkus-openshift-test-suite, is the fix ok to go? @mjurc

mjurc (approved):
This looks good to me. Thanks for the fix!

PR #326, opened 2021-08-20
Upgrade to 2.1.3.Final and fix 999-SNAPSHOT

Sgitario:
The Fabric8 API brings another breaking change, so I had to drop some nice information about why a resource is not loaded.

### DIFF
--- a/common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
+++ b/common/src/main/java/io/quarkus/ts/openshift/common/util/AwaitUtil.java
@@ -11,8 +11,6 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.fabric8.kubernetes.client.Handlers;
-import io.fabric8.kubernetes.client.ResourceHandler;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.quarkus.ts.openshift.app.metadata.AppMetadata;
 import io.quarkus.ts.openshift.common.DefaultTimeout;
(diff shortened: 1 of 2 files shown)


rsvoboda:
Please file issue on Fabric8 side to fix the regression or to add suitable alternative which could be used here.

I would comment out the removed lines ++ add the link to GH issue.

Sgitario:
> Please file issue on Fabric8 side to fix the regression or to add suitable alternative which could be used here.
> 
> I would comment out the removed lines ++ add the link to GH issue.

They removed this feature on purpose to reduce the size of the final jar: https://github.com/fabric8io/kubernetes-client/pull/3327
++ they encounter some issues with custom resources

rsvoboda:
So there is no replacement for you to use to get the details? If so I would still fille the GH issue for them to provide some suggestions. They need to see that there is some impact of their actions.

The change is already in 2.2 branch - https://github.com/quarkusio/quarkus/blob/2.2/bom/application/pom.xml#L173

What are the end user implications? Should be migration guide for Quarkus 2.2 updated?

Sgitario:
> So there is no replacement for you to use to get the details? If so I would still fille the GH issue for them to provide some suggestions. They need to see that there is some impact of their actions.
> 
> The change is already in 2.2 branch - https://github.com/quarkusio/quarkus/blob/2.2/bom/application/pom.xml#L173
> 
> What are the end user implications? Should be migration guide for Quarkus 2.2 updated?

PR updated

rsvoboda (approved):

Sgitario:
> So there is no replacement for you to use to get the details? If so I would still fille the GH issue for them to provide some suggestions. They need to see that there is some impact of their actions.
> 
> The change is already in 2.2 branch - https://github.com/quarkusio/quarkus/blob/2.2/bom/application/pom.xml#L173
> 
> What are the end user implications? Should be migration guide for Quarkus 2.2 updated?

https://github.com/quarkusio/quarkus/issues/19533

PR #356, opened 2021-12-13
Build Quarkus main using profile quickly, enable relocations

rsvoboda:
Build Quarkus main using profile quickly, enable relocations


rsvoboda:
Relocations are not available for org.apache.activemq:artemis-server / org.apache.activemq:artemis-amqp-protocol
We use them in tests, these deps were defined in quarkus-bom before extensions move to quarkiverse

jsmrcka (approved):

PR #455, opened 2022-10-27
Move to io.quarkiverse.artemis groupId

rsvoboda:
Move to io.quarkiverse.artemis groupId


rsvoboda:
https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/455 needs to get in before this one

PR #484, opened 2023-02-07
Move to Quarkus 3.0.0.Alpha3

rsvoboda:
Move to Quarkus 3.0.0.Alpha3

- switch to jakarta packages
- drop of Artemis modules as the extension is not Jakarta ready


rsvoboda:
Updated, missed infinispan troubles as one needs to pass `-Dinclude.datagrid` param.

QPID fails were present because `quarkus-core-only` was setting old version and using different version parameter name. That was deleted as we are not using relocations for qpid because relocation points to pretty old version.

PR #490, opened 2023-03-09
Use hibernate-orm.database.orm-compatibility for Hibernate 6

rsvoboda:
Use hibernate-orm.database.orm-compatibility for Hibernate 6


jsmrcka (approved):
LGTM

